PR #1, opened 2021-03-04
ENT-4116 : Get app running at port 8735

binodpant:
- [X] npm ci and npm run build and npm start, npm test should work locally, do not fail
- [X] app should be up and running at port 8735 locally
- [X] comes with default auth (sso) for now but we will have to fix that up since this app is not going to require auth for users but it still needs to use getAuthenticatedClient eventually to make api calls etc


.env.development
```diff
@@ -0,0 +1,21 @@
+PORT=8735
```
binodpant, line 1: I had to edit this to setup the port

.env.development
```diff
@@ -0,0 +1,21 @@
...
+NODE_ENV='development'
+ACCESS_TOKEN_COOKIE_NAME='edx-jwt-cookie-header-payload'
+BASE_URL='http://localhost:8735'
```
binodpant, line 4: also note this is updated from the template app to match this app

.gitignore
```diff
@@ -0,0 +1,65 @@
+# local dev related
+# https://github.com/edx/frontend-build#local-module-configuration-for-webpack
+.env.private
```
binodpant, line 3: note: .env.privatee won't work yet until we upgrade frontend-build, but that may come with some eslint surprises I think

src/index.jsx
```diff
@@ -0,0 +1,38 @@
...
+  ReactDOM.render(<ErrorPage message={error.message} />, document.getElementById('root'));
+});
+
+initialize({
+  messages: [
+    appMessages,
+    headerMessages,
+    footerMessages,
+  ],
+});
```
taliaedX, line 38: For auth bypassing, can't we just use this? https://github.com/edx/frontend-app-learner-portal-enterprise/blob/44cd9aa88c16c930e3fbf0e0ae222fa60e511533/src/index.jsx#L43
It's also mentioned in the parent ticket ENT-4113.

I agree we will have to figure out separate auth for the API calls, but not sure that is relevant for simply accessing the page and if this flag is all we need, we can add it now no?

binodpant: good call, looks like based on this we may want to se both to false (although it says the first is false by default)

https://github.com/edx/frontend-platform/blob/93ff0609c1de2a8b11a276472e167c48c981f3b2/src/initialize.js#L194

binodpant: and I am puzzled now since both are supposed to be false by default

taliaedX: oh yeah good catch. Not sure why we had to specifically call it out in the learner portal then. Hmmmm

openedx.yaml
```diff
@@ -0,0 +1,9 @@
+# This file describes this Open edX repo, as described in OEP-2:
+# http://open-edx-proposals.readthedocs.io/en/latest/oeps/oep-0002.html#specification
+
+oeps: {}
+openedx-release:
+  # The openedx-release key is described in OEP-10:
+  #   https://open-edx-proposals.readthedocs.io/en/latest/oep-0010-proc-openedx-releases.html
+  # The FAQ might also be helpful: https://openedx.atlassian.net/wiki/spaces/COMM/pages/1331268879/Open+edX+Release+FAQ
+  ref: master
```
nedbat: This MFE will be used by Open edX installations?

src/index.jsx
binodpant: yeah I am still confused though since the app does log me in automatically via sso
so if these flags are off by default why does it do that? maybe there is more to it

binodpant: So confirmed that after logging out, we can still see the contents of the page, meaning log in is not required! so we are all set for this PR. when we start making api calls we will have to see if something else needs to be setup

openedx.yaml
binodpant: hmm, wish I knew this answer. Thanks for raising this! This is an enterprise app for use by external customers without need to log in to edX. I do see we have openedx.yaml files in other enterprise MFEs but I am not sure if it is by design. I may not have enough context to answer this so sourcing @georgebabey for this one

edx-abolger: @nedbat this MFE could be used by open edX installations if they also have edx-enterprise installed, run the enterprise_catalog endpoint, and have configured Algolia as their SaaS catalog index search service. So in theory possible but I am not aware of anyone actually doing it.

edx-abolger: Since the main purpose of this MFE here is to provide a clean UX for Enterprise customers to determine what courses are included in edX's "Subscriptions" and "Online Campus Essentials" B2B products, and that data comes from our database and changes based on business agreements, it's not entirely clear what a community use case would do with this MFE.

binodpant: @nedbat after chatting with George and Adrienne, it feels like we do not want to include this MFE in openedx as of yet. So that would translate to removing this file I presume. We can always have a considered discussion on whether/when it makes sense to do so. Let me know if you have any flags/concerns!

docs/how_tos/i18n.rst
```diff
@@ -0,0 +1,5 @@
+####################
+React App i18n HOWTO
+####################
+
+This document has moved to the frontend-platform repo: https://github.com/edx/frontend-platform/blob/master/docs/how_tos/i18n.rst
```
taliaedX, line 5: why is this doc even included in the template (I assume it was templated) when it just points to a different repo? lol

taliaedX (approved):
It's a shell and thus there's not too much to worry about here. One doc in the template seems kind of useless and I'm unclear on how the transifex stuff is going to work, but we can deal with that later.

docs/how_tos/i18n.rst
binodpant: yeah right? I think there is going be a bit of cleanup potentially in here, I decided I will leave this here for now

openedx.yaml
nedbat: No, you don't remove openedx.yaml.  All of our repos should have this file.  To keep the repo out of the open edx releases, you would remove the openedx-release key.

Help me figure out what we can put into this file to make its use more apparent.

PR #2, opened 2021-03-08
ENT-4125 tabular display of catalog

binodpant:
JIRA: https://openedx.atlassian.net/browse/ENT-4125
Introduces data table into the enterpriseCatalogs component, runs a query with no filters against the catalog index

Tested from staging algolia

things to fix before Merge:

- [x] Need paragon update once it releases from ENT-4290
- [X] styling changes part of this PR or no, did requisite styles

looks like this right now

Screen Shot 2021-03-10 at 4 19 25 PM

<hr/>

Screen Shot 2021-03-10 at 4 19 18 PM

Screen Shot 2021-03-10 at 4 21 39 PM

Screen Shot 2021-03-10 at 4 23 53 PM


src/components/LoadingMessage/index.jsx
```diff
@@ -0,0 +1,23 @@
...
+const LoadingMessage = (props) => {
+  const { className } = props;
+  return (
+    <div
+      className={classNames(
+        'loading d-flex',
+        className,
+      )}
+    >
+      Loading...
```
raq929: This loading module is *terrible*, let's not re-use it. In other places in learner-portal we're using `react-skeleton`, it'll create a much nicer loading experience.

src/components/NotFoundPage.jsx
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+import { Container } from '@edx/paragon';
+import { Helmet } from 'react-helmet';
+import { PAGE_TITLE } from '../constants';
+
+const NotFoundPage = () => (
+  <Container size="lg" className="mt-3">
+    <Helmet title={PAGE_TITLE} />
+    <div className="text-center py-5">
+      This content is not available
```
raq929: Check with UX to be sure this is the text we want.

src/components/NotFoundPage.jsx
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+import { Container } from '@edx/paragon';
+import { Helmet } from 'react-helmet';
+import { PAGE_TITLE } from '../constants';
+
+const NotFoundPage = () => (
+  <Container size="lg" className="mt-3">
+    <Helmet title={PAGE_TITLE} />
```
raq929: This should be the `PageWrapper` component.

src/components/PageWrapper.jsx
```diff
@@ -0,0 +1,17 @@
...
+import { Container } from '@edx/paragon';
+import { Helmet } from 'react-helmet';
+
+import { PAGE_TITLE } from '../constants';
+
+// eslint-disable-next-line react/prop-types
+const Wrapper = ({ children }) => (
+  <Container size="lg" className="mt-3">
+    <Helmet title={PAGE_TITLE} />
+    <div className="text-center py-5">
```
raq929: I think it's unlikely that we'll want `text-center` on an entire page?

src/index.jsx
```diff
@@ -30,6 +28,15 @@ subscribe(APP_INIT_ERROR, (error) => {
 });
 
 initialize({
+  handlers: {
+    config: () => {
+      mergeConfig({
```
raq929, line 33: Yes, awesome, this is definitely the way to be doing this. 
A part of this ticket should also be making a pr on `edx-internal` to add the algolia variables there.

src/store.js
```diff
@@ -0,0 +1,19 @@
+import { applyMiddleware, createStore } from 'redux';
```
raq929: Let's *not* add redux into this project. We don't need it. We can use the `useReducer` hook if we need a reducer, and we can use React context if we need a store.

raq929 (approved):

src/store.js
binodpant: yeah agreed I had started with a store and now I don't want it

src/index.jsx
binodpant: filed: https://openedx.atlassian.net/browse/ENT-4289 for that

binodpant: will take care of it before merging this one, it's very simple and small

src/store.js
binodpant: I remove the store which is not used anyway, I could not remove `react-redux` itself as a dep due to peer dep but yeah it does not use redux store directly anymore

src/components/PageWrapper.jsx
binodpant: yes removed this, don't need on entire page and in fact looks better without it

src/components/NotFoundPage.jsx
binodpant: ran by Joe, updated text

src/index.jsx
binodpant: actually no need to hold up this pr for the ENT-4289 work, that will happen once Talia deploys MFE, and merging this in won't deploy it anywhere yet anyway

.gitignore
```diff
@@ -1,7 +1,7 @@
 # local dev related
 # https://github.com/edx/frontend-build#local-module-configuration-for-webpack
 .env.private
-module.config.js
+module.config.js*
```
raq929, line 4: Why the * ?

package.json
```diff
@@ -37,21 +37,29 @@
...
+    "algoliasearch": "4.8.5",
     "babel-polyfill": "6.26.0",
+    "classnames": "^2.2.6",
     "prop-types": "15.7.2",
     "react": "16.12.0",
     "react-dom": "16.12.0",
-    "react-redux": "7.1.3",
+    "react-helmet": "6.1.0",
+    "react-instantsearch-dom": "6.8.3",
+    "react-loading-skeleton": "^2.2.0",
```
raq929: This will likely get autofixed if we have the dependency bot, but we shouldn't have the `^` anymore.

src/components/PageWrapper.jsx
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
+import { Container } from '@edx/paragon';
+import { Helmet } from 'react-helmet';
+
+import { PAGE_TITLE } from '../constants';
+
+// eslint-disable-next-line react/prop-types
+const Wrapper = ({ children }) => (
+  <Container size="lg" className="mt-3">
```
raq929: What happens on a small screen?

src/components/catalogs/CatalogSearchResults.jsx
```diff
@@ -0,0 +1,127 @@
...
+
+const CatalogSearchResults = ({
+  searchResults,
+  // algolia recommends this prop instead of searching
+  isSearchStalled,
+  searchState,
+  error,
+}) => {
+  if (isSearchStalled) {
+    return (<Skeleton height={25} />);
```
raq929: I think we probably want more than this? Like a skeleton of a table?
Can you put up a screen shot of what this looks like currently?

src/components/catalogs/CatalogSearchResults.jsx
```diff
@@ -0,0 +1,127 @@
...
+  }
+  if (error) {
+    return (
+      <StatusAlert
```
raq929: I think there's a warning about the StatusAlert in the tests in AdminPortal. I think if we're adding it here, we should fix the warnings.

src/components/catalogs/CatalogSearchResults.jsx
```diff
@@ -0,0 +1,127 @@
...
+      Header: TABLE_HEADERS.subject,
+      accessor: 'subjects[0]',
+    },
+    {
+      Header: TABLE_HEADERS.partner,
+      accessor: 'partners[0].name',
+    },
+  ], []);
+
+  const calculatePage = () => {
```
raq929: If you are defining it this function here inside the component, the `useMemo` below is kind of useless. This will always register as a new function and the code inside the `useMemo` will always run.
I'd suggest either moving the definition outside the component, which I like because that makes it testable, or just not memoizing, because this isn't a high-cost operation, so memoizing doesn't seem like a necessary optimization.

src/components/catalogs/CatalogSearchResults.jsx
```diff
@@ -0,0 +1,127 @@
...
+
+  const tableData = useMemo(() => searchResults?.hits || [], [searchResults?.hits]);
+  return (
+    <>
+      <div>
+        <DataTable
+          columns={columns}
+          data={tableData}
+          itemCount={searchResults?.nbHits}
+          isSelectable
```
raq929: Does this table actually need to be selectable? Are people going to perform bulk actions on it? (I don't remember the spec.)

src/components/catalogs/EnterpriseCatalogs.jsx
```diff
@@ -0,0 +1,32 @@
+import React from 'react';
+import {
+  Configure, InstantSearch,
+} from 'react-instantsearch-dom';
+
+import { SearchData, SearchHeader } from '@edx/frontend-enterprise';
+
+import Wrapper from '../PageWrapper';
```
raq929: Let's call it `PageWrapper`; imports get really confusing if we rename them; people used to React will expect this to have the same name as it's file name.

src/components/catalogs/EnterpriseCatalogs.jsx
```diff
@@ -0,0 +1,32 @@
...
+
+  return (
+    <Wrapper>
+      <SearchData>
+        <InstantSearch
+          indexName={algoliaIndexName}
+          searchClient={searchClient}
+        >
+          <Configure hitsPerPage={NUM_RESULTS_PER_PAGE} />
+          <div className="search-header-wrapper">
```
raq929: I expect the SearchHeader doesn't currently accept a `className` prop, but for future reference, I'd prefer to pass this in as a prop rather than adding an extra div.

raq929: Also, I'm not seeing this class used?

src/index.scss
```diff
@@ -5,3 +5,10 @@
 
 @import "~@edx/frontend-component-header/dist/index";
 @import "~@edx/frontend-component-footer/dist/footer";
+
+@import "~@edx/frontend-enterprise/src/course-search/styles/index";
+
+.bg-brand-primary {
```
raq929: Please wrap this in your search header class. `bg-brand-primary` is a css class that might be re-used. We don't want this to accidentally override other styles.
Also, please move this css into the `components/catalogs` folder and import it here. I know we do this in admin portal, but in general, css should live with it's components and the `index.scss` file should *only* contain imports.

raq929 (changes requested):
I've got a few review comments, and this needs tests before it can be merged! Looking good, though!

.gitignore
binodpant: ah yeah I found it convenient to often rename my module.config.js to module.config.js.unused or similar rather than delete and forget what I need.. so this helps with that while keeping current behavior

package.json
binodpant: yeah I missed this one but will clean it up here instead of waiting for the tool ty

src/components/PageWrapper.jsx
binodpant: it works just fine, and checked the doc, 
```
<Container size="lg" className="bg-danger-300 mb-4">
  The content in this container won't exceed the large width.
</Container>
```

 it appears the lg is mainly about not extending beyond a certain width

src/components/catalogs/CatalogSearchResults.jsx
binodpant: hmm when I forced it to show always to see what this shows, and right now I. just see this (should I add a paragon loading spinner in side of this here?)

Screen Shot 2021-03-10 at 11 36 44 AM

binodpant: it's just a animated hardly visible gray bar right now

binodpant: ah single line skeleton right now and small height so looks hard to find, what about this?
Screen Shot 2021-03-10 at 11 45 39 AM

src/components/catalogs/CatalogSearchResults.jsx
binodpant: yeah I just saw onClose is a required field somehow.. .perhaps we need to add it even if not used?

binodpant: hmm also per paragon docs, looks like Alert should be now used instead of StatusAlert anyway, will make that change

binodpant: now looks like this 
Screen Shot 2021-03-10 at 12 11 04 PM

src/components/catalogs/CatalogSearchResults.jsx
binodpant: yeah good callout this is leftover from my attempts to debug issue, don't need this useMemo here

binodpant: think this is simpler 

```
  const page = refinementsFromQueryParams.page || (searchState ? searchState.page : 0);
```

src/components/catalogs/CatalogSearchResults.jsx
binodpant: nope no use case right now for selectable

src/components/catalogs/EnterpriseCatalogs.jsx
binodpant: ah yes removing this since this class is not even defined, and it has no visual effect

src/index.scss
binodpant: good suggestion, done!

binodpant:
all feedback applied!

src/components/catalogs/styles/_index.scss
```diff
@@ -0,0 +1,13 @@
+.enterprise-catalogs-header {
```
raq929, line 1: 🎉

src/index.jsx
```diff
@@ -1,26 +1,24 @@
...
+
+import App from './components/app/App';
 
 import appMessages from './i18n';
 
 import './index.scss';
 
 subscribe(APP_READY, () => {
   ReactDOM.render(
-    <AppProvider>
```
raq929, line 19: Are we not using AppProvider because we don't need authentication?

src/components/catalogs/CatalogSearchResults.jsx
raq929: Yeah, that'll do for the moment. 😄  We can run it by UX later.

src/components/catalogs/CatalogSearchResults.jsx
raq929: Throw a top margin on that, and I think we're good.

src/index.jsx
binodpant: oh wait ok I do  use it, just in the <App> component

src/components/catalogs/CatalogSearchResults.jsx
```diff
@@ -10,21 +10,34 @@ import {
...
 };
 
-const CatalogSearchResults = ({
+/**
+ * The core search resultss rendering component.
+ *
+ * Wrapping this in `connectStateResults()` will inject the first few props.
+ *
+ * @param {object} args arguments
+ * @param {object} args.searchResults Results of search (see: `connectStateResults``)
```
binodpant, line 29: I do not like the 'hidden-ness' of props like this that come from connected/HOC components, so I feel better when I doc it like this ...

src/components/catalogs/CatalogSearchResults.jsx
binodpant: done! mt-2 for both alerts

src/components/catalogs/CatalogSearchResults.jsx
raq929: This is what PropTypes is for? I don't see the point of further documenting it like this, but up to you.

src/components/catalogs/CatalogSearchResults.jsx
```diff
@@ -10,21 +10,34 @@ import {
...
+ * @param {object} args.error Error with `message` field if available (see: `connectStateResults``)
+ * @param {object} args.paginationComponent Defaults to <SearchPagination> but can be injected
+ */
+export const CoreCatalogSearchResults = ({
   searchResults,
   // algolia recommends this prop instead of searching
   isSearchStalled,
   searchState,
   error,
+  paginationComponent: PaginationComponent = SearchPagination,
```
raq929: At edx we generally use `defaultProps` rather than doing it like this, but I don't really have an issue with it.

src/components/catalogs/CatalogSearchResults.jsx
```diff
@@ -107,6 +120,7 @@ CatalogSearchResults.propTypes = {
   searchState: PropTypes.shape({
     page: PropTypes.number,
   }).isRequired,
+  paginationComponent: PropTypes.node.isRequired,
 };
 
-export default connectStateResults(CatalogSearchResults);
+export default connectStateResults(CoreCatalogSearchResults);
```
raq929: Please use `BaseCatalogSearchResults` instead of `Core`, to be conform with standards used in other repos.

src/components/NotFoundPage.jsx
```diff
@@ -0,0 +1,9 @@
+import React from 'react';
+import PageWrapper from './PageWrapper';
+
+const NotFoundPage = () => (
```
raq929, line 5: It's a very simple page right now, but let's add a test for this component.

src/components/PageWrapper.jsx
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
+import { Container } from '@edx/paragon';
+import { Helmet } from 'react-helmet';
+
+import { PAGE_TITLE } from '../constants';
+
+// eslint-disable-next-line react/prop-types
+const Wrapper = ({ children }) => (
```
raq929, line 10: This component also needs tests.

src/components/app/App.jsx
```diff
@@ -0,0 +1,22 @@
...
+
+import { EnterpriseCatalogs } from '../catalogs';
+import NotFoundPage from '../NotFoundPage';
+
+export default function App() {
+  return (
+    <AppProvider>
+      <Header />
+      <Switch>
+        <PageRoute exact path="/" component={EnterpriseCatalogs} />
```
raq929: I'd like to start this repo off right with testing, and test our routes. 
Example here: https://techdoma.in/react-js-testing/how-to-test-routes-react-router-dom-in-react-using-jest-1

src/components/catalogs/CatalogSearchResults.jsx
```diff
@@ -0,0 +1,126 @@
...
+export const CoreCatalogSearchResults = ({
+  searchResults,
+  // algolia recommends this prop instead of searching
+  isSearchStalled,
+  searchState,
+  error,
+  paginationComponent: PaginationComponent = SearchPagination,
+}) => {
+  if (isSearchStalled) {
+    return (<Skeleton className="m-1 loading-skeleton" height={25} count={5} />);
```
raq929: We should test that this is displayed when the `isSearchStalled` is true.

src/components/catalogs/CatalogSearchResults.jsx
```diff
@@ -0,0 +1,126 @@
...
+      <Alert className="mt-2" variant="warning">
+        {NO_DATA_MESSAGE}
+      </Alert>
+    );
+  }
+
+  const { refinementsFromQueryParams } = useContext(SearchContext);
+  const columns = useMemo(() => [
+    {
+      Header: TABLE_HEADERS.courseName,
```
raq929, line 73: We should test that the correct table headers are displayed in the document.

src/components/catalogs/EnterpriseCatalogs.jsx
```diff
@@ -0,0 +1,30 @@
...
+} from 'react-instantsearch-dom';
+
+import { SearchData, SearchHeader } from '@edx/frontend-enterprise';
+
+import PageWrapper from '../PageWrapper';
+import { NUM_RESULTS_PER_PAGE } from '../../constants';
+import CatalogSearchResults from './CatalogSearchResults';
+import { useAlgoliaIndex } from './data/hooks';
+
+export default function EnterpriseCatalogs() {
```
raq929, line 13: I don't think this component needs a lot in terms of testing, but we should test it if at all possible.

src/components/catalogs/data/hooks.jsx
```diff
@@ -0,0 +1,21 @@
+/* eslint-disable import/prefer-default-export */
+import { useMemo } from 'react';
+
+import algoliasearch from 'algoliasearch/lite';
+
+import { getConfig } from '@edx/frontend-platform';
+
+const useAlgoliaIndex = () => {
```
raq929, line 8: Please test this hook.

raq929 (commented):
TL; DR More tests please!
What you've got is looking good!

src/components/catalogs/EnterpriseCatalogs.jsx
raq929: I couldn't find a great way to mock Algolia, though, so... this is the only one I'm kind of willing to skip if necessary.

src/components/catalogs/CatalogSearchResults.jsx
binodpant: yeah some of this may come down to taste / prefs too totally, I just ilke being able to read in the function doc what things are for etc, proptypes tells me the type but not the purpose esp. from 'magical' places like connected components

src/components/catalogs/CatalogSearchResults.jsx
binodpant: this is probably just my years of OOP muscle brain haha. `Base` makes me think of inheritance or components down from there.. and there are not. But I also see consistency as a good way to help understand so happy to change this instead of adding yet another prefix term!

src/components/app/App.jsx
binodpant: yeah there are more than one route, so testing routes is a good idea

src/components/catalogs/EnterpriseCatalogs.jsx
binodpant: this one could get a bit tricky but let me see what can be done.. if seems too much work maybe I will let it be, but yeah worth a try

src/index.jsx
raq929: If you're using it, it belongs here; it needs to wrap the entire app, not just the component. Things will get *very* confused if it's not at the top level.

src/components/catalogs/EnterpriseCatalogs.jsx
binodpant: managed to setup a test thatt renders the main component (except the AppProvider part since I want to mock the data that goes into it), using this doc as guide: https://testing-library.com/docs/example-react-router/

src/components/app/App.jsx
binodpant: tested!

src/components/catalogs/CatalogSearchResults.jsx
binodpant: tested!

raq929 (approved):
I love it! :shipit:

PR #3, opened 2021-03-25
feat: update frontend-enterprise

raq929:
Adds skills dropdown!
I forgot to add it to the screen shot but this does modify the url correctly.
Note: screen shot does not have the edx theme
Screen Shot 2021-03-25 at 11 43 54 AM


binodpant (approved):
wohoo you're on a roll! this is exciting to have here, sure Joe will love this

PR #4, opened 2021-03-26
Add Hero and Call to Action

raq929:
This adds the Hero and Call to Action at the top of the page.
It refactors / changes some styles and updates frontend-enterprise to take advantage of the "default" variation of the SearchHeader.

**Note**
The text varies from the mock-up, I discussed those changes with Jenn.

Full screen
Screen Shot 2021-03-30 at 12 54 58 PM

Mobile:
Screen Shot 2021-03-30 at 12 55 35 PM


package.json
```diff
@@ -34,7 +34,7 @@
     "url": "https://github.com/edx/frontend-app-enterprise-public-catalog/issues"
   },
   "dependencies": {
-    "@edx/brand": "npm:@edx/brand-openedx@1.1.0",
```
binodpant, line 37: so yeah for this project do we also not follow the same pattern of openedx default, and override brand during build process?

src/styles/_variables.scss
```diff
@@ -0,0 +1,5 @@
+// Hero
+$hero-max-height: 350px;
+
+// banners (Hero, callToAction)
+$banner-min-height: 298px;
```
binodpant, line 5: for this I have edited my VSCode editor to auto insert newlines at end which helps ...  

```
"files.insertFinalNewline": true
```

src/setupTest.js
```diff
@@ -1 +1,9 @@
 import 'babel-polyfill';
+
+// Fail tests if there are console errors
+const { error } = global.console;
```
binodpant, line 4: nice!

src/index.scss
```diff
@@ -1,11 +1,15 @@
...
+@import "~@edx/brand/paragon/variables.scss";
 @import "@edx/paragon/scss/core/core.scss";
-@import "@edx/brand/paragon/overrides.scss";
+@import "~@edx/brand/paragon/overrides.scss";
 
 @import "~@edx/frontend-component-header/dist/index";
 @import "~@edx/frontend-component-footer/dist/footer";
 
-@import "~@edx/frontend-enterprise/src/course-search/styles/index";
+@import "~@edx/frontend-enterprise";
```
binodpant, line 9: love this change too!

src/components/hero/Hero.jsx
```diff
@@ -0,0 +1,41 @@
...
+import PropTypes from 'prop-types';
+import SmallHeroImage from '../../assets/hero-image-cluster-1x.png';
+import LargeHeroImage from '../../assets/hero-image-cluster-2x.png';
+
+const imageAltText = 'People learning and performing highly skilled tasks';
+
+export const FirstWordHighlighter = ({ text }) => {
+  const words = text.split(' ');
+  const firstWord = words.shift();
+  return <span className="word-splitter"><div className="highlighted">{`${firstWord} `}</div><div>{words.join(' ')}</div></span>;
```
binodpant: nit: my slight pref for multilining this jsx but nothing big, as long as eslint is happy

src/components/catalogs/tests/CatalogSearchResults.test.jsx
```diff
@@ -129,10 +129,9 @@ describe('Main Catalogs view works as expected', () => {
       <SearchDataWrapper>
         <BaseCatalogSearchResults
           paginationComponent={PaginationComponent}
-          searchResults
+          searchResults={searchResults}
           isSearchStalled
-          searchState
-          error
+          searchState={{ page: 1 }}
```
binodpant, line 134: ah did not realize if there was a default page setting expectation for this component ... I thought it went to 1 by default.. but having it here is explicit and better regardless

binodpant (approved):
this looks awesome to me, tests are good too! made a few minor comments and suggestions

src/components/catalogs/tests/CatalogSearchResults.test.jsx
raq929: This was setting `searchState` to true, which caused a PropTypes error in the tests. I have set our tests to fail on that kind of error in the future.

raq929:
Now with text highlighting
Screen Shot 2021-03-30 at 2 23 28 PM

src/styles/_variables.scss
raq929: YES THANK YOU

package.json
raq929: I think not? I just updated to version 2.0. Oh, it seems like I changed from brand-openedx to @edx/brand-edx.
I don't know what is right, to be honest.

binodpant: @adamstankiewicz can confirm but I think for other projects we have it setup so we use the openedx one by default, and have `something` (I think part of the MFE deployment setup somewhere) do the replacement of brand in the build process, to use the brand-edx one.

the template app also comes with the openedx brand by default: https://github.com/edx/frontend-template-application/blob/master/package.json#L37

For this project, I am not super clear on if that is needed. Do we just default to the brand-edx branding or use openedx by default? Will there be openedx users of this project, will drive that decision I think, for this one perhaps check with Joe C as well?

binodpant:
looks nice, clever highlight work! just added a question about the brand setup but otherwise looking great! now that I know of the tool, maybe also try ARC toolkit against this just for fun…

binodpant (approved):
just that brand question to follow up on but otherwise lgtm

package.json
raq929: I'm gonna merge as-is and let @alex-sheehan-edx (tech lead) figure it out with Joe/Adam as needed.

PR #12, opened 2021-04-01
Internationalization Setup

taliaedX:
### Description
This configures wrappers and message formats (for currently written messages) to support translations from transifex.
It also adds the transifex config.
The only remaining bits would be to actually create the pipeline jobs and run them which we decided was out of scope for ENT-4258.

#### Please help me oh god
There is one test failing and I cannot for the life of me get it to work, please halp.
It is the CatalogSearchResults test.

#### Locale: EN
Screen Shot 2021-04-01 at 4 14 24 PM
#### Locale: RU (with local Russian translation file - purposely not fully translated to clearly show it is populated from the files)
Screen Shot 2021-04-01 at 4 20 16 PM

### Live Testing:
If you want to see this truly in action, do the following:
- update one of the language files in src/i18n/messages/<locale>.json
  - You need to add JSON keys that equal the ids for `FormattedMessages` and/or the <Component>.messages.js files.
  - Example of an updated `ru.json` file with all of the catalog page text translated:
```
{
    "catalogPage.hero.text": "Индивидуальное обучение для вашей команды",
    "catalogPage.hero.highlight": "Индивидуальное обучение",
    "catalogPage.cta.button.text": "Запланировать звонок",
    "catalogPage.cta.business.link": "для бизнеса",
    "catalogPage.cta.campus.link": "для учебных заведений",
    "catalogPage.cta.text": "Изучите исчерпывающие каталоги курсов, созданные {businessLink} и {campusLink}, или поработайте с представителем edX, чтобы настроить решение в соответствии с вашими уникальными потребностями."
}
```
- You would then need to add a language to your Chrome browser
  - (Settings > Advanced > Languages > Add Languages > Pick one that we support (such as French, Russian, etc.)
  - move the language to the top (click the 3 dots next to the language name > "Move to top")
  - Open an incognito window
  - navigate to the page (`localhost:8735`). Voila.

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,6 +15,7 @@
     "i18n_extract": "BABEL_ENV=i18n fedx-scripts babel src --quiet > /dev/null",
     "is-es5": "es-check es5 ./dist/*.js",
     "lint": "fedx-scripts eslint --ext .js --ext .jsx .",
+    "lint:fix": "fedx-scripts eslint --ext .js --ext .jsx . --fix",
     "snapshot": "fedx-scripts jest --updateSnapshot",
     "start": "fedx-scripts webpack-dev-server --progress",
     "test": "fedx-scripts jest --coverage --passWithNoTests"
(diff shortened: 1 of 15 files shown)


package.json
```diff
@@ -15,6 +15,7 @@
     "i18n_extract": "BABEL_ENV=i18n fedx-scripts babel src --quiet > /dev/null",
     "is-es5": "es-check es5 ./dist/*.js",
     "lint": "fedx-scripts eslint --ext .js --ext .jsx .",
+    "lint:fix": "fedx-scripts eslint --ext .js --ext .jsx . --fix",
```
raq929, line 18: 🎉

raq929 (approved):
Looks good to me! 
And I'm super glad we're doing this now, it'll give us a good template to work from as we add text to the page.

PR #35, opened 2021-04-06
feat: Marketing url configured on CTA link

taliaedX:
This replaces the static url defined on the CatalogPage with a variable generated from the config at run time.

The variable chosen here - `ENTERPRISE_MARKETING_URL` - is actually already in use with the exact values we want in edx-internal: https://github.com/edx/edx-internal/blob/master/frontends/common/stage_config.yml#L37
(Yes the same value exists in the common prod config as well). So we don't even need to open a PR there for this change either.

There are no visible UI changes here, hence no screenshots.

I also updated the env.development file just so devs copying it for use in a private.py know what the names of the variables are without having to look at index.jsx or wherever else we end up sticking things.


edx-abolger (approved):
LGTM!

alex-sheehan-edx (approved):
I feel like error handling for cases when this isn't filled isn't necessary. ✅  LGTM!

PR #36, opened 2021-04-13
Update images to use srcset

raq929:
Uses srcset and sets a size for the images for better calculations of when to use hi and lo-res images.


src/components/hero/Hero.jsx
```diff
@@ -4,21 +4,37 @@ import {
...
 const Hero = ({ intl, text, highlight }) => (
   <section className="hero">
     <Container size="lg" className="hero__content">
       <h1 className="display-3"><Highlighted text={text} highlight={highlight} /></h1>
       <div>
-        <Large>
-          <Image className="hero__image" src={SmallHeroImage} alt={intl.formatMessage(messages['hero.image.alt'])} />
+        <Large className={IMAGE_WRAPPER_CLASS}>
+          <Image
+            className="hero__image"
```
alex-sheehan-edx: is there a reason why `hero__image-wrapper` was put in a constant but not `hero_image`?

alex-sheehan-edx (approved):
One small, non-blocking question. Otherwise LGTM! 👍

src/components/hero/Hero.jsx
raq929: Oh, because I was going to use it instead of `div` which is not very precise, but the class doesn't get passed through, so... I should remove this.

PR #37, opened 2021-04-13
Set the minimal header variable

raq929:
AUTHN_MINIMAL_HEADER is needed on dev, but MINIMAL_HEADER is needed on prod. Included here mostly to remind people that it's needed on prod.


.env.development
```diff
@@ -19,7 +19,9 @@ REFRESH_ACCESS_TOKEN_ENDPOINT='http://localhost:18000/login_refresh'
 SEGMENT_KEY=''
 SITE_NAME='edX'
 USER_INFO_COOKIE_NAME='edx-user-info'
+ENTERPRISE_MARKETING_URL='http://example.com'
 ALGOLIA_APP_ID=''
 ALGOLIA_SEARCH_API_KEY=''
 ALGOLIA_INDEX_NAME=''
-ENTERPRISE_MARKETING_URL='http://example.com'
```
taliaedX, line 25: confused by the order change here, but yeah otherwise a-okay.

README.rst
```diff
@@ -9,6 +9,7 @@ Introduction
 This application is a public facing catalog page for use by edX consumers to find courses in catalogs before deciding to enroll.
 
 The dev server is running at `http://localhost:8735 <http://localhost:8735>`_.
+The staging server is running at `https://explore-catalog.stage.edx.org/`_.
```
taliaedX, line 12: ooo good add.

.env.development
```diff
@@ -19,7 +19,9 @@ REFRESH_ACCESS_TOKEN_ENDPOINT='http://localhost:18000/login_refresh'
...
+AUTHN_MINIMAL_HEADER=true
```
taliaedX, line 26: do we also need this specified for prod/stage (I just noted in the edx-internal PR, because I looked at that first derp)

taliaedX (approved):
LGTM. May want to make sure if we need the AUTHN_MINIMAL_HEADER value or not and if so, we need to make sure it's in that edx-internal pr.

PR #38, opened 2021-04-14
FEAT: Adding catalog selection UX cards

alex-sheehan-edx:
screen shot:
image

mobile screenshot:
image

[jira](https://openedx.atlassian.net/browse/ENT-4412)

Adds catalog selection cards to the public catalog page. 

Features:
- only one card is selectable at a time
- no selected cards will result in filtering nothing, displaying the entire index
- selecting a card will update the url's query params
- loading the page with a query param of a particular card's id will result in that card being checked
- selecting cards will act in the same way as selecting options from the data table's drop down menu
- internationalization should be applied to all parts of the cards

TODO's:
After this is implemented, we will need to replace the `env` catalog uuid references for edx-internal configurable UUIDs associated with the 3 catalog queries that we want for the cards.


.env.development
```diff
@@ -23,3 +23,6 @@ ALGOLIA_APP_ID=''
 ALGOLIA_SEARCH_API_KEY=''
 ALGOLIA_INDEX_NAME=''
 ENTERPRISE_MARKETING_URL='http://example.com'
+EDX_FOR_BUSINESS_UUID='5afdb7d0-e7b6-4e81-ad4b-b02b8036b7e2'
```
taliaedX: I'd leave the development ones blank actually. I don't know if these are the prod UUIDs I passed you and Dusenbery yesterday or not, but whether it's those or your own local catalogs, they won't exist for someone else locally.

taliaedX: ie this should look the same as the `.env` file changes.

.env.test
```diff
@@ -17,3 +17,7 @@ REFRESH_ACCESS_TOKEN_ENDPOINT='http://localhost:18000/login_refresh'
 SEGMENT_KEY=''
 SITE_NAME='edX'
 USER_INFO_COOKIE_NAME='edx-user-info'
+EDX_FOR_BUSINESS_UUID='ayylmao'
+EDX_FOR_ONLINE_EDU_UUID='foo'
+EDX_ONLINE_ESSENTIALS_UUID='bar'
+ENTERPRISE_MARKETING_URL='http://apptest.com'
```
taliaedX, line 20: I'm guessing this and the fact I see these same config variables being re-initialized in the test app setup is where the questions of "what is the point of `.env.test`" were coming from in the frontend channel. I can't actually tell from the conversation in that channel why this is still here though. For what it's worth, the learner portal frontend doesn't even have an env.test, so I'm 99.99% sure we could just throw it out.

src/components/catalogSelectionCard/CatalogSelectionCard.jsx
```diff
@@ -0,0 +1,50 @@
...
+  useContext,
+} from 'react';
+import PropTypes from 'prop-types';
+import { SearchContext, deleteRefinementAction, setRefinementAction } from '@edx/frontend-enterprise';
+
+import {
+  Card, Form,
+} from '@edx/paragon';
+
+export const QUERY_UUID_REFINEMENT = 'enterprise_catalog_query_uuids';
```
taliaedX: this exact verbiage is used in the CatalogPage searchFacetFilter's attribute field. Should we move this to a general constants file and re-use a bit more?

taliaedX: Also used in the catalog selection deck test

src/components/catalogs/EnterpriseCatalogs.jsx
```diff
@@ -23,16 +23,14 @@ export default function EnterpriseCatalogs() {
...
-            <InstantSearch
-              indexName={algoliaIndexName}
-              searchClient={searchClient}
-            >
-              <Configure hitsPerPage={NUM_RESULTS_PER_PAGE} />
-              <div className="enterprise-catalogs-header"><SearchHeader variant="default" /></div>
-              <CatalogSearchResults />
-            </InstantSearch>
-          </SearchData>
+          <InstantSearch
```
taliaedX, line 26: I haven't looked too much into these individual components, so this is entirely just dumb curiosity, but what is the reasoning for removing the SearchData wrapper?

taliaedX (approved):
There's some redundancies that could do with deletion and/or making constants, but otherwise LGTM!

.env.test
alex-sheehan-edx: While testing with Lael, we found that this file wasn't actually being used... fun configuration messes! This was left in as a mistake (not even needed) as we have tests working without it.

taliaedX: yeah let's throw it out to avoid someone thinking it's actually useful in the future. 😛

src/components/catalogSelectionCard/CatalogSelectionCard.jsx
alex-sheehan-edx: done!

src/components/catalogs/EnterpriseCatalogs.jsx
alex-sheehan-edx: The search data wrapper had to be moved to the [catalog page component](https://github.com/edx/frontend-app-enterprise-public-catalog/pull/38/files#diff-951cff911bae1321495d1fbed1100618082ad1195377e882510cae205837b9aaR48-R52) to include the cards within the context! Lael would be much more qualified to speak towards this, but from our pair programing my understanding is that the wrapper defines the context in which the data table will be able to use information/filters, so to allow the cards to filter the data table, they needed to be included along with the enterprise catalog component.

raq929: Secretly the cards set algolia filters and the data table just displays the result, but yeah, we needed the cards to be within the SearchData context because SearchData does a lot of nice handling of the query params for us.

raq929 (approved):

PR #39, opened 2021-04-15
feat: :sparkles: Add tracking events ENT-4383 Part 1 just CTA button

binodpant:
This PR is not the entirety of the ticket, just one event to get things rolling in small pieces.

Tracking events included in this PR:
- [x] Tracking event for cta button click only

The other tracking event for search I will do in a separate PR just to keep things moving (that may actually need frontend-enterprise plus this portal PR)

I think page visits is already covered by PageRoute

ENT-4126

### DIFF
--- a/src/index.jsx
+++ b/src/index.jsx
@@ -1,5 +1,6 @@
 import 'babel-polyfill';
 
+import { sendPageEvent } from '@edx/frontend-platform/analytics';
 import {
   APP_INIT_ERROR, APP_READY, subscribe, initialize, mergeConfig,
 } from '@edx/frontend-platform';
@@ -21,6 +22,7 @@ subscribe(APP_READY, () => {
     <App />,
     document.getElementById('root'),
   );
+  sendPageEvent();
 });
 
 subscribe(APP_INIT_ERROR, (error) => {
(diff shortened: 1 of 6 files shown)


src/index.jsx
```diff
@@ -21,6 +22,7 @@ subscribe(APP_READY, () => {
     <App />,
     document.getElementById('root'),
   );
+  sendPageEvent();
```
taliaedX: given we are already mocking this in a test prior to this pr here: https://github.com/edx/frontend-app-enterprise-public-catalog/blob/main/src/components/app/tests/App.test.jsx#L11
I'm not sure this explicitly needs to be called here. Did you find this line necessary to send events?

taliaedX (approved):
Aside from what seems like an unnecessary method call, this looks fine to me.

src/index.jsx
taliaedX: Seems like PageRoute calls it which is a couple levels under the overarching App layer and I do see other frontend repos explicitly calling this method. I guess the question is: What are we tracking at App rendering that wouldn't already be covered underneath by the `EnterpriseCatalogsApp` level?

binodpant: ah wait good catch yes we odn't need this esp since pageroute call it already per route

PR #43, opened 2021-04-23
fix: adding blank target to force hyperlinks to open in new tabs

alex-sheehan-edx:
https://edx.github.io/paragon/components/hyperlink/


binodpant (approved):

edx-abolger:
@alex-sheehan-edx  and @binodpant  I am posting this here for your knowledge: this change should not have been approved as is.  This is a very common HTML issue you should both be aware of as developers in the future: Please update as noted in the following article: https://perishablepress.com/fix-blank-target-vulnerability/.  It's not a security hole, but it is more like a commonly known future vulnerability, so think of adding the tag as kind of an extra fail-safe, future-proof.

PR #44, opened 2021-04-27
ENT-4126 add segment tracking

binodpant:
Tested by sending a round of local events to staging segment source.

also renamed tracking name for cta button to be consistent with the ones used in frontend-enterprise using `explore-catalog`

in Segment, verified that:
* page load event works
* cta button click tracks
* search via clicking on button tracks with info like this

Screen Shot 2021-04-27 at 3 10 07 PM


binodpant:
@alex-sheehan-edx perhaps you already found a fix in frontend-enterprise for this? this PR now also fails the same way with 'acorn' is-es5 checks

package.json
```diff
@@ -72,12 +71,13 @@
...
+    "@testing-library/user-event": "13.1.3",
     "codecov": "3.7.2",
     "es-check": "5.1.4",
     "glob": "7.1.6",
     "husky": "6.0.0",
     "jest": "24.9.0",
     "reactifex": "1.1.1"
-  }
+  },
+  "sideEffects": false
```
binodpant: per https://edx-internal.slack.com/archives/C01GNDN7TN3/p1619573759059600?thread_ts=1619556598.047800&cid=C01GNDN7TN3

package.json
```diff
@@ -35,18 +35,17 @@
     "url": "https://github.com/edx/frontend-app-enterprise-public-catalog/issues"
   },
   "dependencies": {
-    "@edx/brand": "npm:@edx/brand-edx.org@2.0.0",
+    "@edx/brand": "npm:@edx/brand-openedx@1.1.0",
```
taliaedX, line 38: not sure we want this change.

binodpant: yeah I am still torn on this one. On one hand @adamstankiewicz informed me that all other enterprise MFEs do this so openedx users can see the default theme. in stage/prod the npm aliasing will actually replace this theme with the enterprise brand theme. however the story on this app vs openedx was far from clear when we last had a slack chat. so for now I decided it may be best to make this consistent with the others? thoughts? This change does mean that locally seeing the other theme requires an extra npm install, but that is now standard with all the MFEs

binodpant: https://github.com/edx/edx-internal/blob/master/frontends/frontend-app-explore-catalog/prod_config.yml#L14

package.json
```diff
@@ -35,18 +35,17 @@
...
     "@edx/frontend-enterprise": "5.6.1",
     "@edx/frontend-platform": "1.8.4",
-    "@edx/paragon": "^14.6.1",
+    "@edx/paragon": "14.6.1",
     "@fortawesome/fontawesome-svg-core": "1.2.34",
     "@fortawesome/free-brands-svg-icons": "5.11.2",
     "@fortawesome/free-regular-svg-icons": "5.11.2",
     "@fortawesome/free-solid-svg-icons": "5.11.2",
     "@fortawesome/react-fontawesome": "0.1.14",
-    "@testing-library/user-event": "^13.1.2",
```
alex-sheehan-edx, line 49: 👍

src/constants.js
```diff
@@ -8,6 +8,9 @@ export const NUM_RESULTS_PER_PAGE = 40;
  * Spec: https://openedx.atlassian.net/wiki/spaces/AN/pages/2624455676/Enterprise+public+catalog+Event+Design
  */
 export const PRODUCT_EVENTS = {
-  ctaButtonPressed: 'edx.enterprise.public_catalog.cta_button_pressed',
+  ctaButtonPressed: 'edx.enterprise.explore-catalog.cta_button_pressed',
 };
+export const TRACKING_APP_NAME = 'explore-catalog';
+// end: tracking related
```
alex-sheehan-edx, line 14: Is this comment necessary?

alex-sheehan-edx: if it is, perhaps going above the line would be more clear

alex-sheehan-edx (approved):
one potential cruft related comment but otherwise LGTM. PSA you will need to manually deploy the prod gocd pipeline when you want this deployed

src/constants.js
binodpant: ya just to clarify what it’s for

binodpant: it marks thecend of the tracking config section

package.json
taliaedX: ahhhh didn't realize it gets overwritten. That's fine then. I thought it was unintentional

PR #47, opened 2021-04-30
fix: changing marketing url from shared URL to catalog specific one

alex-sheehan-edx:
NOTE:
-This needs to be merged after the new config value is added to edx internal
-I'm not sure `.env.test` does _**anything**_ so I didn't bother adding the new config value to it.

`ENTERPRISE_MARKETING_URL` is a common config value shared among MFE's, we want Catalog to have a hubspot specific one to be labeled: `HUBSPOT_MARKETING_URL`


.env.development
```diff
@@ -30,3 +29,4 @@ EDX_FOR_ONLINE_EDU_UUID=''
 EDX_ONLINE_ESSENTIALS_UUID=''
 FEATURE_PROGRAM_TITLES_FACET=True
 FEATURE_LANGUAGE_FACET=True
+HUBSPOT_MARKETING_URL='http://example.com'
```
taliaedX, line 32: Oh nice! I was going to suggest the name be Hubspot related since I assumed this specific site was lol

src/components/app/tests/App.test.jsx
```diff
@@ -22,7 +22,7 @@ jest.mock('react-instantsearch-dom', () => ({
 
 const mockConfig = () => (
   {
-    ENTERPRISE_MARKETING_URL: 'http://apptest.com',
+    HUBSPOT_MARKETING_URL: 'http://bobsdooremporium.com',
```
taliaedX, line 25: I see what you did here.

taliaedX (approved):
LGTM! As you yourself have noted, make sure internal goes well!

PR #49, opened 2021-06-04
fix: Algolia - query with filters to get accurate pill counts

taliaedX:
When we do not filter our initial query to Algolia, facet counts are based on the total records in the index while the result set is inherently deduped.
This means that if we get 500 hits from our search we may have pill facet values of 47000.

If you filter on your initial query, the facet counts seem to be based on the filtered result set (which is also already deduped). 
Note: I am still waiting on Algolia support to explain why this is exactly, but regardless....

Thus, I have added functionality to filter by all of our catalog queries prior to a user selecting one.
If they do select a catalog query, it swaps to that as normal.

From the UI perspective there should be absolutely no change and mostly functionally there is no change, except that we actually get useful counts.

#### Algolia FormData before this change:
Screen Shot 2021-06-04 at 3 25 17 PM

#### Algolia data and view after this change:
Screen Shot 2021-06-04 at 3 24 18 PM

(If you couldn't tell, before, we had no filters, after we filter on using all 3 catalogs)

### DIFF
--- a/src/components/catalogs/EnterpriseCatalogs.jsx
+++ b/src/components/catalogs/EnterpriseCatalogs.jsx
@@ -8,11 +8,11 @@ import { FormattedMessage } from '@edx/frontend-platform/i18n';
 import PageWrapper from '../PageWrapper';
 import { NUM_RESULTS_PER_PAGE } from '../../constants';
 import CatalogSearchResults from './CatalogSearchResults';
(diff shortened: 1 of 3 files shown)


alex-sheehan-edx (approved):
Some slight complexity added for when we inevitably remove the essentials card/catalog filter from the public catalog but no matter. This is really well done! 🎉 LGTM ✅

PR #50, opened 2021-06-09
maint: separating out and installing catalog-search 

alex-sheehan-edx:
[jira](https://openedx.atlassian.net/browse/ENT-4500)

By separating the catalog search package from frontend enterprise, this PR accomplishes two things:

- pulls in changes to add New Relic synthetic attributes for monitoring and alerting
- removes unnecessary logistration related packages from public catalog


binodpant:
would be nice to see some testing notes on how you verified this, but the changes themselves look fine to me based on what I know! Will leave for Adam to approve, I am also warming up to the lerna system so presuming Adam and you have done due diligence on testing. Would still drop notes if you have any workflow wise in case there is a similar upcoming change someone else works on (even if just a link). Otherwise lgtm!

adamstankiewicz (approved):
Looks great!

alex-sheehan-edx:
> would be nice to see some testing notes on how you verified this, but the changes themselves look fine to me based on what I know! Will leave for Adam to approve, I am also warming up to the lerna system so presuming Adam and you have done due diligence on testing. Would still drop notes if you have any workflow wise in case there is a similar upcoming change someone else works on (even if just a link). Otherwise lgtm!

@binodpant happy to add testing notes to the description, but TL:DR you should really just be able to pull the branch, `npm install`, start the webpack server and view the new relic synthetic data attributes on the input, submit and cancel elements. That will confirm that it's successfully using the newest version of  `frontend-enterprise-catalog-search`. You can also confirm that `frontend-enterprise` is no longer listed in the package json after you npm install

alex-sheehan-edx:
@binodpant I believe Adam has done a wonderful write up on lerna and deploying changes on frontend-enterprise. This PR is mostly just putting into effect his work of splitting `frontend-enterprise` into three packages.

https://github.com/edx/frontend-enterprise/blob/42bdd2ce061ebf1f788d5684ca6a709fe1be9671/docs/how_tos/resolving_publishing_issues_with_lerna.rst

There is also one small note which is locally you may have the openedx branding instead of the one that we use on prod- you simply need to swap out `"npm:@edx/brand-openedx@1.1.0"` within your package.json with `"npm:@edx/brand-edx.org@2"` and rerun `npm i` as well as restart the webpack server

binodpant:
ok great! looks like the mentioned doc plus your notes here are more than enough to cover my thoughts! thanks

PR #52, opened 2021-06-14
Fix: Force calculation of facets (the pill counts) to occur after results (aka deduplication) returned.

taliaedX:
Algolia support has indicated that facets are always calculated before deduplication unless using `facetingAfterDistinct`. The reason us just filtering seemed to work was because the values we were filtering on happened to not have any duplicated results. (Thus we should really apply this setting to the learner portal as well to ensure no weird counts there).

This PR just adds that `facetingAfterDistinct` flag to the configuration for our queries.

### Behavior on Prod:
Screen Shot 2021-06-14 at 12 28 15 PM

### Behaviour locally with this change:
Screen Shot 2021-06-14 at 12 26 04 PM

(Ignore any weird UX differences, it's due to the brand themes being different)

### DIFF
--- a/src/components/catalogs/EnterpriseCatalogs.jsx
+++ b/src/components/catalogs/EnterpriseCatalogs.jsx
@@ -27,7 +27,7 @@ export default function EnterpriseCatalogs() {
             indexName={algoliaIndexName}
             searchClient={searchClient}
           >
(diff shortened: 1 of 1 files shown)


alex-sheehan-edx (approved):
LGTM ✅  thx for getting this out so quickly

PR #54, opened 2021-06-15
ENT-4637 Add tag to stop the robots.

edx-abolger:
Note: this solution can't be used if you also use robots.txt.


binodpant (approved):
lgtm, assuming we do not use robots.txt to block any crawling, and we should not have any other html pages to apply this tag to

PR #55, opened 2021-06-16
fix: revert changes from ent-4539 - Show all courses

taliaedX:
The initial attempt to fix the facet counts in ent-4539 made the conscious change to pre-filter our result set. We filtered to the 3 catalog options listed in the explore-catalog and selecting any 1 just isolated the results down to that specific catalog.

The behaviour product actually wants is to show *all* courses (not just the ones in those 3 catalogs) prior to filtering.
With the implementation of ENT-4664, we can do this and *still* fix the underlying issue with facet counts (since 4664 contains the actual fix)

If you look here: https://github.com/edx/frontend-app-enterprise-public-catalog/pull/49
You can see I'm just reverting what I added. We had to do some string concatenation to be able to query multiple catalogs at once.

### Before this Change (Production index values):
Screen Shot 2021-06-16 at 3 06 12 PM

### After this Change:
Screen Shot 2021-06-16 at 2 55 28 PM

### DIFF
--- a/src/components/catalogs/EnterpriseCatalogs.jsx
+++ b/src/components/catalogs/EnterpriseCatalogs.jsx
@@ -8,11 +8,11 @@ import { FormattedMessage } from '@edx/frontend-platform/i18n';
 import PageWrapper from '../PageWrapper';
 import { NUM_RESULTS_PER_PAGE } from '../../constants';
 import CatalogSearchResults from './CatalogSearchResults';
-import { useAlgoliaIndex, useDefaultSearchFilters } from './data/hooks';
+import { useAlgoliaIndex } from './data/hooks';
 
 export default function EnterpriseCatalogs() {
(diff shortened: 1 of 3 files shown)


raq929 (approved):

src/components/catalogs/EnterpriseCatalogs.jsx
```diff
@@ -27,7 +27,7 @@ export default function EnterpriseCatalogs() {
             indexName={algoliaIndexName}
             searchClient={searchClient}
           >
-            <Configure hitsPerPage={NUM_RESULTS_PER_PAGE} filters={filters} facetingAfterDistinct />
+            <Configure hitsPerPage={NUM_RESULTS_PER_PAGE} facetingAfterDistinct />
```
taliaedX, line 30: facetingAfterDistinct is all we actually need to fix facet counts.

PR #106, opened 2021-09-01
build: :arrow_up: Upgrade to frontend-enterprise 2.0.7 and fix breaking changes as a result.

edx-abolger:
This PR upgrades the frontend-enterprise dependency to 2.0.7 and contains the fixes needed to mitigate the breaking changes from 2 major version bumps. This is necessary to pull in changes to fix ENT-3656

_Because_ this involved breaking changes from dependencies, I also upgraded the package version to 1.0.0, not that anything depends on it right now.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@edx/frontend-app-enterprise-public-catalog",
-  "version": "0.1.0",
+  "version": "1.0.0",
   "description": "Frontend application template",
   "repository": {
     "type": "git",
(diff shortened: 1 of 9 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@edx/frontend-app-enterprise-public-catalog",
-  "version": "0.1.0",
+  "version": "1.0.0",
```
taliaedX, line 3: are the changes here big enough for a major version change? (I'm bad at versioning, but it feels like mostly minor changes)

taliaedX (approved):
Looks fine to me. I have one question around versioning, but not a blocker.

package.json
edx-abolger: Yes, the breaking change is because frontend-enterprise bump breaks changes that change prop names, stopping any backwards compatibility

PR #115, opened 2021-09-07
feat: adding a price column to the search data results

alex-sheehan-edx:
[Jira](https://openedx.atlassian.net/browse/ENT-4604)

description:
now that first enrollable seat price is available on algolia, we can add a la carte pricing column to the public catalog

image

current questions: is there a placeholder that we want to display for when there isn't a price available? I don't know how common it's gonna be to have an unavailable price.

### DIFF
--- a/src/components/catalogSearchResults/CatalogSearchResults.jsx
+++ b/src/components/catalogSearchResults/CatalogSearchResults.jsx
@@ -41,6 +41,7 @@ export const BaseCatalogSearchResults = ({
   const TABLE_HEADERS = {
     courseName: intl.formatMessage(messages['catalogSearchResults.table.courseName']),
     partner: intl.formatMessage(messages['catalogSearchResults.table.partner']),
+    price: intl.formatMessage(messages['catalogSearchResults.table.price']),
     catalogs: intl.formatMessage(messages['catalogSearchResults.table.catalogs']),
   };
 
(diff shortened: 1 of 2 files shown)


src/components/catalogSearchResults/CatalogSearchResults.messages.js
```diff
@@ -11,6 +11,11 @@ const messages = defineMessages({
...
     description: 'The partner institution providing/authoring the course (ie Harvard, MIT, etc.)',
   },
+  'catalogSearchResults.table.price': {
+    id: 'catalogSearchResults.table.price',
+    defaultMessage: 'A la carte course price',
+    description: 'Table column A La Carte price for the course',
+  },
   'catalogSearchResults.table.catalogs': {
     id: 'catalogSearchResults.table.catalogs',
     defaultMessage: 'Associated Catalogs',
```
kiram15: lol this is my bad, but can you change it to "Associated catalogs" so the capitalization is the same for all the columns?

src/components/catalogSearchResults/CatalogSearchResults.jsx
```diff
@@ -92,6 +93,10 @@ export const BaseCatalogSearchResults = ({
       Header: TABLE_HEADERS.partner,
       accessor: 'partners[0].name',
     },
+    {
+      Header: TABLE_HEADERS.price,
+      accessor: 'first_enrollable_paid_seat_price',
```
kiram15, line 98: I think the prices need a preceding dollar sign. Also, do we account for people who are paying with different currencies?

kiram15 (commented):
Super quick fixes but looks good other than that!

src/components/catalogSearchResults/CatalogSearchResults.jsx
alex-sheehan-edx: as far as I know, we just deal in dollars. Appending something to the accessor might be a pain, let me look into it.

alex-sheehan-edx: done 😄

kiram15 (approved):
👍 🚀

PR #119, opened 2021-09-13
maint: switching query uuid facet for title

alex-sheehan-edx:
Switching public catalog over to title facets instead of uuid's. Note prod needs to be reindexed before this is merged.


binodpant (approved):
lgtm!

PR #124, opened 2021-09-24
ENT-4973: Altering modal's scroll, adding course availability dates

kiram15:
Screenshots also available[ in the ticket ](https://openedx.atlassian.net/secure/RapidBoard.jspa?rapidView=630&projectKey=ENT&modal=detail&selectedIssue=ENT-4973)for clarity, cleared by UX (Jenn)

### DIFF
--- a/src/components/catalogCourseInfoModal/CatalogCourseInfoModal.jsx
+++ b/src/components/catalogCourseInfoModal/CatalogCourseInfoModal.jsx
@@ -15,9 +15,9 @@ import { Launch } from '@edx/paragon/icons';
 import messages from './CatalogCourseInfoModal.messages';
 
 function formatDate(start, end) {
-  const startDate = (new Date(start.startDate)).toLocaleDateString();
-  const endDate = (new Date(end.endDate)).toLocaleDateString();
-  return ((` ${startDate} - ${endDate}`));
+  const startDate = (new Date(start)).toLocaleDateString();
+  const endDate = (new Date(end)).toLocaleDateString();
+  return `${startDate} - ${endDate}`;
 }
 
 const CatalogCourseInfoModal = ({
@@ -92,7 +92,7 @@ const CatalogCourseInfoModal = ({
             </p>
             <p className="course-info-description-subtitle">
               {intl.formatMessage(messages['catalogCourseInfoModal.availabilityMessage'])}
-              {formatDate({ startDate }, { endDate })}
+              {formatDate(startDate, endDate)}
             </p>
             {/* eslint-disable-next-line react/no-danger */}
             <div dangerouslySetInnerHTML={{ __html: courseDescription }} />
--- a/src/components/catalogCourseInfoModal/CatalogCourseInfoModal.jsx
+++ b/src/components/catalogCourseInfoModal/CatalogCourseInfoModal.jsx
@@ -14,6 +14,12 @@ import { injectIntl, intlShape } from '@edx/frontend-platform/i18n';
 import { Launch } from '@edx/paragon/icons';
 import messages from './CatalogCourseInfoModal.messages';
 
+function formatDate(start, end) {
+  const startDate = (new Date(start.startDate)).toLocaleDateString();
+  const endDate = (new Date(end.endDate)).toLocaleDateString();
+  return ((` ${startDate} - ${endDate}`));
+}
+
 const CatalogCourseInfoModal = ({
   intl,
   isOpen,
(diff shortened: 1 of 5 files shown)


src/components/catalogCourseInfoModal/CatalogCourseInfoModal.jsx
```diff
@@ -38,55 +47,56 @@ const CatalogCourseInfoModal = ({
...
+                  {intl.formatMessage(messages['catalogCourseInfoModal.educationBadge'])}
+                </Badge>
+              )}
+            </div>
+            <p className="course-info-description-title">
+              {intl.formatMessage(messages['catalogCourseInfoModal.courseDescriptionTitle'])}
+            </p>
+            <p className="course-info-description-subtitle">
+              {intl.formatMessage(messages['catalogCourseInfoModal.availabilityMessage'])}
+              {formatDate({ startDate }, { endDate })}
```
binodpant: this is a curious input form? should it be `formatDate(startDate, endDate)` instead? alternatively you can also go with `formatDate({startDate, endDate}) as long as you destructure in the function def.

src/components/catalogCourseInfoModal/CatalogCourseInfoModal.jsx
```diff
@@ -14,6 +14,12 @@ import { injectIntl, intlShape } from '@edx/frontend-platform/i18n';
 import { Launch } from '@edx/paragon/icons';
 import messages from './CatalogCourseInfoModal.messages';
 
+function formatDate(start, end) {
+  const startDate = (new Date(start.startDate)).toLocaleDateString();
```
binodpant: based on the proptypes, looks like `start` and `end` should be strings and not objects?

src/components/catalogCourseInfoModal/CatalogCourseInfoModal.jsx
```diff
@@ -14,6 +14,12 @@ import { injectIntl, intlShape } from '@edx/frontend-platform/i18n';
...
+  const endDate = (new Date(end.endDate)).toLocaleDateString();
+  return ((` ${startDate} - ${endDate}`));
```
binodpant: is this extra leading space actually needed or?

binodpant: I think the outer parens are superflous, maybe just try :

```
return `${startDate} - ${endDate}`
```

src/components/catalogCourseInfoModal/CatalogCourseInfoModal.jsx
alex-sheehan-edx: I was testing with her- printing out the variables interestingly yielded dicts, maybe the proptype should be changed?

src/components/catalogCourseInfoModal/CatalogCourseInfoModal.jsx
```diff
@@ -38,55 +47,56 @@ const CatalogCourseInfoModal = ({
...
         className="course-info-modal"
       >
-        <ModalDialog.Hero className="course-info-hero">
-          <ModalDialog.Hero.Background backgroundSrc={bannerImageUrl} />
-        </ModalDialog.Hero>
-
-        <ModalDialog.Header className="course-info-modal-header">
-
+        <ModalDialog.Body className="full-body">
+          <ModalDialog.Hero>
```
alex-sheehan-edx, line 51: love that you can just throw this into the body

src/components/catalogCourseInfoModal/CatalogCourseInfoModal.jsx
alex-sheehan-edx: 😄 I had spotted the same thing before seeing your comment- this is why https://github.com/edx/frontend-app-enterprise-public-catalog/pull/124/files#diff-caee6ed0461fad8f827d7e53ce7b4f859c9ff39f02d45323ed7d6d8bda2d1cd6R18 is an object

alex-sheehan-edx (approved):
I agree with @binodpant's comments, + I think you have params to tweak but otherwise this looks great 😄

src/components/catalogCourseInfoModal/CatalogCourseInfoModal.jsx
kiram15: I have the extra leading space just to give it a space after the dash in availability, but I can put that in the message if that makes more sense!

PR #129, opened 2021-10-14
feat: adding in custom functionality for ICE

kiram15:
[ENT-5041](https://openedx.atlassian.net/browse/ENT-5041) details the custom work that Joe requested after a customer call with ICE. They wanted customization to hide the price column, replacing it with an available date column. They also wanted the opportunity to hide the catalog selection cards. The screenshot below shows the public catalog with both url parameters applied. 

The link on prod will read as `https://explore-catalog.edx.org/?enterprise_catalog_query_titles=Education&availability=Available%20Now&availability=Upcoming&hide_price=true&hide_cards=true`

Screen Shot 2021-10-14 at 8 09 49 AM

### DIFF
--- a/src/components/catalogs/styles/_enterprise_catalogs.scss
+++ b/src/components/catalogs/styles/_enterprise_catalogs.scss
@@ -58,7 +58,7 @@
 .pricing-data-header {
   position: absolute;
   right: 5%;
-  top: 52%;
+  top: 66%;
 }
 
 .pricing-data-title {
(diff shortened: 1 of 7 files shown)


src/components/catalogs/styles/_enterprise_catalogs.scss
```diff
@@ -58,7 +58,7 @@
 .pricing-data-header {
   position: absolute;
   right: 5%;
-  top: 52%;
```
kiram15, line 61: This is a quick formatting thing that I'm sneaking in here because it's been bothering me, not directly related to this ticket.

johnnagro (approved):
lgtm ⛵

PR #147, opened 2021-11-03
build: use shared browserslist config

adamstankiewicz:
* Removes custom `browserslist` configuration in favor of using a [shared configuration](https://github.com/edx/browserslist-config).
* Removes `is-es5` check in CI since ES5 was only needed for IE 11 support which is dropped. The supported browsers defined by the shared configuration all [natively support ES6](https://caniuse.com/?search=es6).

This change reduces the resultant asset bundle size.

# For all changes

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,7 +24,5 @@ jobs:
       run: npm run build
     - name: i18n_extract
       run: npm run i18n_extract
-    - name: is-es5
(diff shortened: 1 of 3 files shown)


binodpant:
@adamstankiewicz is this PR blocked by the npm upgrades? I saw a passing note elsewhere but figured I would check specifically for this PR if that applies. If yes, cool as well! just wanted to check in

PR #157, opened 2021-11-19
feat: allow card view when cardView setting is on

binodpant:
# Feature controlled card view vs list view toggling

The feature switch in config is off by default but when on, it allows switching between card and list views

Card view is default if feature is on, otherwise, table view is default (for now)

Current does not do the following:

- support usage within Data tAble's control bar, but it's fine for this UI since this one does not yet use Data Table bulk actions.
- use a Paragon component for toggle. It's a local toggle component used for now with a boolean switch to switch views, and also feature controlled with a flag
- finalize card look feel and behavior, that will be done in a separate ticket

# For all changes

https://user-images.githubusercontent.com/528166/142936365-1d58a333-68ba-43c8-80ec-d5b3587f7752.mov

This actually shows the buttons final look (I force hover style when you click on the iconbutton, for now). It gets us closer but not 100% compliance with Figma (we can't do 100% due to IconButton limitation)

https://user-images.githubusercontent.com/528166/142935649-17af8e9f-42e0-4cde-9647-94eaa0a82033.mov


src/components/catalogSearchResults/CatalogSearchResults.jsx
```diff
@@ -6,12 +7,14 @@ import React, {
 import PropTypes from 'prop-types';
 import { connectStateResults } from 'react-instantsearch-dom';
 import {
-  Badge, DataTable, Alert, Button, useToggle,
+  Badge, DataTable, Alert, Button, useToggle, CardView, Card, Icon, IconButton,
```
kiram15: nit: Organize these in alphabetical order and/or put them all on their own line for readability

src/components/catalogSearchResults/CatalogSearchResults.jsx
```diff
@@ -109,6 +140,33 @@ export const BaseCatalogSearchResults = ({
...
+        onClick={() => { setCardView(true); }}
+        variant={cardView ? 'dark' : 'light'}
+        className="mr-2"
+      />
+      <IconButton
+        src={ListView}
+        iconAs={Icon}
+        alt="Card view"
+        onClick={() => { setCardView(false); }}
+        variant={!cardView ? 'dark' : 'light'}
```
kiram15: The button should stay dark to show as selected (I think you're already working on this)

src/components/catalogSearchResults/CatalogSearchResults.jsx
```diff
@@ -31,6 +34,34 @@ function formatDate(courseRun) {
...
+
+const CourseCard = ({ className, original }) => {
+  const { title, card_image_url, partners } = original;
+
+  return (
+    <Card className={className}>
+      <Card.Img
+        variant="top"
+        src={card_image_url}
+        style={{ width: '48vh', height: '26vh' }}
```
kiram15: According to Figma, there should be 4 cards as a max when the screen can accommodate, right now I'm only seeing 3

.nvmrc
```diff
@@ -1 +1 @@
-12.22.7
+v12
```
kiram15, line 1: What is the purpose of the change in this file?

kiram15 (changes requested):

src/components/catalogSearchResults/CatalogSearchResults.jsx
```diff
@@ -109,6 +140,33 @@ export const BaseCatalogSearchResults = ({
...
   const [startDate, setStartDate] = useState();
   const [endDate, setEndDate] = useState();
+  const [cardView, setCardView] = useState(true);
+
+  // TODO: Feature control for Card view. Remove once cards are finalized
+  const config = getConfig();
+  const cardViewEnabled = config.FEATURE_CARD_VIEW_ENABLED === 'true';
+
+  // TODO: local view toggle compoent. To be replaced by IconButtonToggle from Paragon
+  const ViewToggle = () => (
```
kiram15: We should probably also have tests for the toggle element to validate the change (which I also think you're working on atm)

.nvmrc
binodpant: ah this is just for future looking npm upgrade friendliness. Actually v12 will make the most recent 12 npm be used. But after my npm upgrade work is done this will instead change to v16. It does actually have no impact currently unless you run `nvm use` locally, then it will ask for installing the latest v12 node if not already

src/components/catalogSearchResults/CatalogSearchResults.jsx
binodpant: agreed! I am now working on some tests

src/components/catalogSearchResults/CatalogSearchResults.jsx
binodpant: ah good callout! I have to revisit how I did this, I will see if I can make 4 cards show per row instead. I don;t know if we had a min width requirement for these course cards though, mybe simply reducing maxWidth will work let's see

src/components/catalogSearchResults/CatalogSearchResults.jsx
binodpant: yup I managed to make it stay dark. Looks like I am missing making it actually color invert when selected like in the figma spec so will check that out

binodpant: as discussed ,I got close, but not 100% per figma due to iconbutton limitation. we can revisit once we  work on the IconButton.Toggle component

kiram15 (approved):
LGTM! 👍 🚀

src/components/catalogSearchResults/CatalogSearchResults.jsx
```diff
@@ -208,7 +296,18 @@ export const BaseCatalogSearchResults = ({
           pageSize={searchResults?.hitsPerPage || 0}
         >
           <DataTable.TableControlBar />
-          <DataTable.Table />
+          { cardViewEnabled && cardView ? (
+            <CardView
```
binodpant, line 300: this uses the excellent column sizes api exposed by cardview to show requisite cols per row, given different screen sizes

PR #162, opened 2021-11-29
feat: add skills section ENT-4971

binodpant:
Shows upto 5 skills taken from the 'skill_names' key in the algolia response

There is a horizontal scrollbar issue but it was there before this PR and is also in prod so not related to this PR

Screen Shot 2021-11-29 at 2 25 20 PM

# For all changes

### DIFF
--- a/src/components/catalogCourseInfoModal/CatalogCourseInfoModal.test.jsx
+++ b/src/components/catalogCourseInfoModal/CatalogCourseInfoModal.test.jsx
@@ -88,4 +88,23 @@ describe('Course info modal works as expected', () => {
     expect(screen.queryByText('A la carte')).toBeInTheDocument();
     expect(screen.queryByText('Business')).not.toBeInTheDocument();
   });
+  test('modal displays upto 5 skills list', () => {
+    const defaultPropsCopy = {
+      ...defaultProps,
+      skillNames: [...Array(20).keys()].map(i => `skill-${i}`),
+    };
+
+    render(
+      <IntlProvider locale="en">
+        <CatalogCourseInfoModal {...defaultPropsCopy} />
+      </IntlProvider>,
+    );
+    expect(screen.queryByText('skill-0')).toBeInTheDocument();
+    expect(screen.queryByText('skill-1')).toBeInTheDocument();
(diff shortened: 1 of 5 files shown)


src/components/catalogs/styles/_enterprise_catalogs.scss
```diff
@@ -49,10 +49,8 @@
   height: 70%;
   max-height: 90px;
   background-color: $light-300;
-  box-shadow: 0 1px 1px rgba(128,128,128,.1),
-              0 1px 1px rgba(128,128,128,.1),
-              0 2px 2px rgba(128,128,128,.1),
```
kiram15, line 54: Nit: I will say I think this old one might be more readable

src/components/catalogCourseInfoModal/CatalogCourseInfoModal.test.jsx
```diff
@@ -88,4 +88,23 @@ describe('Course info modal works as expected', () => {
     expect(screen.queryByText('A la carte')).toBeInTheDocument();
     expect(screen.queryByText('Business')).not.toBeInTheDocument();
   });
+  test('modal displays upto 5 skills list', () => {
```
kiram15, line 91: My only additional test is to add one if no skills are not found? But not necessary

kiram15 (approved):
A couple nit picks but otherwise LGTM!

src/components/catalogs/styles/_enterprise_catalogs.scss
binodpant: hmm yeah thanks for the catch my editor has reformatted automatically somehow, .. can check

src/components/catalogCourseInfoModal/CatalogCourseInfoModal.test.jsx
binodpant: yeah I think it's good to add an explicit check in the other tests or just add one with the skills absent, will do that!

src/components/catalogs/styles/_enterprise_catalogs.scss
binodpant: hmm my editor auto format refused to take me back to this style :(  I am using eslint as my formatter

PR #165, opened 2021-12-03
style fixes ENT-5199

binodpant:
Tested locally by:

```
npm i @edx/brand@npm:@edx/brand-edx.org  # to install the edx brand theme
npm start
```

Fixes included:

[1] : skills display bullet list was wrapping text after certain screen size reduction
Basically changing display:flex to display:block for the element (including the `<ul>`) fixes this issue, it was a mistake to use `display:flex` for the `ul` since we are using block level inline elements for the `<li>` and the `ul` itself should behave like a block level element for this use case to property support the inline block elements. 

[2]: the floating button group now has a slightly higher top padding (using bootstrap `pt-1` to fix the 'too little space' issue from the top of the surrounding div. Looks like this now:

Screen Shot 2021-12-03 at 1 33 47 PM

# For all changes

### For issue [1], after fixes

small screen
small screen example 2


kiram15 (approved):
LGTM!

PR #166, opened 2021-12-03
fix: fix horiz scrollbar and other CSS issues ENT-5085

binodpant:
[1] fix for scrollbar:
    By using hidden, we will not allow scrolling right/left. It was not fully clear to me but it looks like the sentinel divs on top and bottom (the pseudo elements) were responsible for the scrollbars, since none of the content was actually overflowing. However, this avoid the horiz scroll bars in any width

[2] fix for image size shrinkage for partner logo
instead of % (which is based on parent), using 12vw width (which is per viewport instead), and also constrain min width to the 200px size that we typically see on desktop screens. This makes image width more predictable and also no crazy shrinkages. See attached.

[3] fix for checkbox size oddity. This involved making flex-grow and flex-shrink to be 0 to disable the resizing, and also a small amount of css cleanup to only style the radio area and rename some css classes

Screen Shot 2021-12-07 at 11 12 44 AM
Screen Shot 2021-12-07 at 11 12 34 AM
Screen Shot 2021-12-07 at 11 12 21 AM
Screen Shot 2021-12-07 at 11 12 10 AM

https://user-images.githubusercontent.com/528166/145067562-fab34826-bf49-49db-baf0-2e2a5f59595c.mov

[4] fix for hero image. 
now don't show at small, xsmall but do show image at medium widths too
also show image at extraextralarge widths (wide screens)

Screen Shot 2021-12-08 at 4 31 00 PM

https://user-images.githubusercontent.com/528166/145288488-2278712a-36e2-4304-8e47-8a0f15cfb9be.mov

Screen Shot 2021-12-08 at 3 11 25 PM

will remove draft status once the other fixes are included/verified

# For all changes

### DIFF
--- a/src/components/catalogs/styles/_enterprise_catalogs.scss
+++ b/src/components/catalogs/styles/_enterprise_catalogs.scss
@@ -11,6 +11,7 @@
 
 .padded-body {
   margin: 180px 0 0 24px;
+  padding: 0em 0.9em;
 }
 
 .course-info-title {
--- a/src/components/catalogs/styles/_enterprise_catalogs.scss
+++ b/src/components/catalogs/styles/_enterprise_catalogs.scss
@@ -6,6 +6,7 @@
 
 .full-body {
   padding: 0px !important;
+  overflow-x: hidden; /* avoid horizontal scrollbar */
 }
 
 .padded-body {
--- a/src/components/catalogs/styles/_enterprise_catalogs.scss
+++ b/src/components/catalogs/styles/_enterprise_catalogs.scss
@@ -11,7 +11,7 @@
 
 .padded-body {
   margin: 180px 0 0 24px;
-  padding: 0em 0.9em;
+  padding: 0em 0.9em 0em 0em;
 }
 
 .course-info-title {
--- a/src/components/catalogs/styles/_enterprise_catalogs.scss
+++ b/src/components/catalogs/styles/_enterprise_catalogs.scss
@@ -46,7 +46,8 @@
   z-index: 1;
   left: 2%;
   top: 230px;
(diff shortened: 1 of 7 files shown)


src/components/catalogs/styles/_enterprise_catalogs.scss
```diff
@@ -6,6 +6,7 @@
 
 .full-body {
   padding: 0px !important;
+  overflow-x: hidden; /* avoid horizontal scrollbar */
```
kiram15, line 9: Could we add additional padding to both skills grey box and course description so that it has the same as on the left?

kiram15 (changes requested):
Once we add in the fix for the hero this looks good!

src/components/catalogs/styles/_enterprise_catalogs.scss
binodpant: sure! how is this ? added 0.9em padding to both right and left


Screen Shot 2021-12-08 at 4 42 54 PM


before is like this

Screen Shot 2021-12-08 at 4 44 14 PM

kiram15: I think it would be better if only added to the right side? Now the left looks like too much - apologies for the nit

kiram15 (approved):
One little change with the padding requested but otherwise LGTM!

src/components/catalogs/styles/_enterprise_catalogs.scss
binodpant: no not at all, thanks for being detailed, I only apply right padding now

PR #178, opened 2021-12-17
feat: adding program card functionality/formatting

kiram15:
# To Test 

Use this link - http://localhost:8735/?content_type=program 
Locally, in CatalogSearchResults.jsx
1. Add import ProgramCard from '../programCard/ProgramCard';
2. Change line 312 to CardComponent={(props) => <ProgramCard {...props} className="program-card" onClick={cardClicked} />}
3. Set FEATURE_PROGRAM_TYPE_FACET=True in your .env.development files
3. Note: Click is not functional in this iteration, that work is detailed in a future ticket 

# For all changes


src/components/PageWrapper.jsx
```diff
@@ -8,7 +8,7 @@ export const DATA_TEST_ID = 'enterprise-catalogs-content';
 
 // eslint-disable-next-line react/prop-types
 const PageWrapper = ({ children, className }) => (
-  <Container size="lg" className={className}>
```
kiram15, line 11: This width mimics B2C so our cards look similar instead of being skinny even in the screens largest width.

src/components/catalogPage/CatalogPage.jsx
```diff
@@ -24,11 +32,13 @@ const CatalogPage = ({ intl }) => {
   const loadedSearchParams = new URLSearchParams(window.location.search);
   let reloadPage = false;
   let hideCards = false;
-  if (config.EDX_ENTERPRISE_ALACARTE_TITLE && (!loadedSearchParams.get(QUERY_TITLE_REFINEMENT))) {
```
kiram15, line 27: This change is made because the programs in staging are not properly associated with the catalogs - https://openedx.atlassian.net/browse/ENT-5249. This logic will be removed when it is fixed.

src/components/programCard/ProgramCard.jsx
```diff
@@ -0,0 +1,106 @@
...
+            className={classNames(
+              'program d-flex justify-content-center align-items-center text-primary-500', 'mb-2',
+            )}
+          >
+            <Icon src={Program} className="badge-icon" />
+            <span className="badge-text"> {program_type} </span>
+          </Badge>
+        </div>
+        <p className="x-small mb-2 mt-2">{makePlural(course_keys.length, 'Course')}</p>
+        {enterprise_catalog_query_titles && (
```
kiram15, line 62: If we don't have the boolean of enterprise_catalog_query_titles, the staging data will error out, but this will be removed when we can guarantee that every program has an associated catalog.

src/components/programCard/ProgramCard.jsx
```diff
@@ -0,0 +1,106 @@
...
+
+import { injectIntl, intlShape } from '@edx/frontend-platform/i18n';
+import messages from './ProgramCard.messages';
+
+function makePlural(num, string) {
+  if (num > 1) { return (`${num} ${string}s`); }
+  return (`${num} ${string}`);
+}
+
+const programCard = ({
```
binodpant: I think it's better to use naming for components : PascalCase, so `ProgramCard` here instead

src/index.scss
```diff
@@ -20,11 +20,18 @@
...
+  margin: 4px;
 }
+
+.bright.badge {
+  color: $black;
+}
+
+.program.badge {
+  font-family: $font-family-monospace;
+}
```
binodpant: we will want to add a newline here

src/components/catalogPage/CatalogPage.jsx
```diff
@@ -1,3 +1,5 @@
+/* eslint-disable camelcase */
```
binodpant: not a must for this PR: But I think we can look at this util from frontend-platform so we can avoid turning off the camel case rule? 

```
import { camelCaseObject } from '@edx/frontend-platform/utils';
```

I see frontend app learner portal uses this util quite a bit 

e.g.,
```
return [camelCaseObject(courseData), fetchError];
```

kiram15: I don't think will work within the way I'm using the variables in proptypes, but I removed it in other places!

src/components/catalogs/styles/_enterprise_catalogs.scss
```diff
@@ -93,22 +92,34 @@
...
 
+// pc - program card
+.program-card {
+  background: $primary-500;
+  border: 0px;
+  color: $white;
+}
+
+.program-title {
+  font-weight: 700;
```
binodpant, line 121: can we use a variable here? I am not sure but we may have access to a proper var from paragon here

binodpant (approved):
synced with Kira offline on open actoin item https://github.com/openedx/frontend-app-enterprise-public-catalog/pull/178/files#r772392267 as well as tested locally out with instructions, properly, cards render and this is featured off so it looks good for this round of work

binodpant:
I am good after minor feedback addressed

PR #180, opened 2021-12-22
feat: use new Data view toggle in datatable ENT-5242

binodpant:
Needs new Paragon release

depends on : https://github.com/edx/paragon/pull/988
since it requires new Paragon release

ENT-5242

# For all changes

### DIFF
--- a/src/components/catalogSearchResults/CatalogSearchResults.jsx
+++ b/src/components/catalogSearchResults/CatalogSearchResults.jsx
@@ -283,8 +283,13 @@ export const BaseCatalogSearchResults = ({
         upcomingRuns={upcomingRuns}
         skillNames={skillNames}
       />
-      { cardViewEnabled && <ViewToggle cardView={cardView} setCardView={setCardView} /> }
       <DataTable
+        dataViewToggleOptions={{
+          isDataViewToggleEnabled: cardViewEnabled,
+          onDataViewToggle: val => setCardView(val === 'card'),
+          togglePlacement: 'left',
+          defaultActiveStateValue: 'card',
+        }}
         columns={columns}
         data={tableData}
         itemCount={searchResults?.nbHits}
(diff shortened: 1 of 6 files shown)


src/components/catalogSearchResults/CatalogSearchResults.jsx
```diff
@@ -299,7 +304,7 @@ export const BaseCatalogSearchResults = ({
         )}
       >
         <DataTable.TableControlBar />
-        { cardViewEnabled && cardView ? (
+        { cardViewEnabled && cardView && (
```
kiram15: Do we still need the boolean of "cardViewEnabled"? That was mostly just used as a psuedo-feature flag to not show the functionality before it was ready, I don't see any reason why we would disable it in the future

binodpant: You are correct. If we want to also enable the feature as part of this PR I can clean this out, thoughts? I was initially going to propose doing that in a separate PR after we validate staging, but I feel now that we might have enough confidence to do so once we look at staging, with this feature flag removed as well

binodpant: alternatively we can leave this in here now, and turn on feature in prod too, and then clean it out later.. upto you!

kiram15: I think we should clean it up now, just because another PR will cause more delays

binodpant: cool let's just take it out of here now, let me do that

src/components/catalogSearchResults/CatalogSearchResults.jsx
```diff
@@ -299,7 +300,7 @@ export const BaseCatalogSearchResults = ({
         )}
       >
         <DataTable.TableControlBar />
-        { cardViewEnabled && cardView ? (
+        { cardView && (
```
adamstankiewicz, line 303: nit: could leave this as a ternary instead of having an explicit `!cardView` case

adamstankiewicz (approved):
LGTM. Just one non-blocking nit/suggestion.

kiram15 (approved):

PR #183, opened 2022-01-06
fix: course search csv download unicode fix

johnnagro:
- follow-on to #182
- force a BOM

ENT-5291


alex-sheehan-edx (approved):
👀 on this one :shipit:

PR #186, opened 2022-01-19
feat: adding programs to explore catalog

kiram15:
# For testing 
To test feature toggle: append `&show_programs=true` to the end of the link

To test actually seeing programs when you toggle the feature: 
- Change line 68 in CatalogPage.jsx to `[...SEARCH_FACET_FILTERS]`
- Comment out line 44-61 in CourseCard.jsx (all the badge code)
- Comment out 257-270 in CatalogSearchResults.jsx and replace with empty div

This PR does NOT include working modal, download button, or error messages, as that work is captured in other tickets. This just creates the foundation for that code to be built upon. 

# For all changes

Screen recording in the comments of the original ticket ([ENT-5206](https://openedx.atlassian.net/browse/ENT-5206))


src/components/catalogSearchResults/CatalogSearchResults.jsx
```diff
@@ -204,9 +211,22 @@ export const BaseCatalogSearchResults = ({
...
 
+  const refinementClick = (content) => {
+    if (content === CONTENT_TYPE_COURSE) {
+      dispatch(setRefinementAction(CONTENT_TYPE_REFINEMENT, [CONTENT_TYPE_COURSE]));
+    } else {
+      dispatch(setRefinementAction(CONTENT_TYPE_REFINEMENT, [CONTENT_TYPE_PROGRAM]));
+    }
+  };
+
+  function cardType(props) {
```
binodpant: nit: should we name this something like `renderCardComponent()` or `renderContentCard()` instead?

src/components/catalogSearchResults/CatalogSearchResults.jsx
```diff
@@ -353,6 +447,8 @@ BaseCatalogSearchResults.propTypes = {
   paginationComponent: PropTypes.func,
   // eslint-disable-next-line react/no-unused-prop-types
   row: PropTypes.string,
+  contentType: PropTypes.string.isRequired,
+  preview: PropTypes.bool,
```
binodpant, line 463: do we plan to keep the preview boolean permanently, is this a feature switch? do we need it to ship this PR?

src/components/catalogs/CatalogSearch.jsx
```diff
@@ -0,0 +1,84 @@
...
+            defaultMessage="Search courses and programs"
+            description="Search dialogue."
+            tagName="h2"
+          />
+          <InstantSearch
+            indexName={algoliaIndexName}
+            searchClient={searchClient}
+          >
+            <div className="enterprise-catalogs-header"><SearchHeader hideTitle variant="default" /></div>
+            {(!contentType || contentType.length === 2) && (
```
binodpant: not clear why we need this check? can we get multiple contentType values in this value from algolia, or none?

src/components/catalogs/CatalogSearch.jsx
```diff
@@ -0,0 +1,84 @@
...
+import { useAlgoliaIndex } from './data/hooks';
+import PageWrapper from '../PageWrapper';
+
+import {
+  CONTENT_TYPE_COURSE, CONTENT_TYPE_PROGRAM, NUM_RESULTS_COURSE, NUM_RESULTS_PROGRAM, NUM_RESULTS_PER_PAGE,
+} from '../../constants';
+import CatalogSearchResults from '../catalogSearchResults/CatalogSearchResults';
+
+export default function CatalogSearch() {
+  const { refinements: { content_type: contentType } } = useContext(SearchContext);
```
binodpant: if contentType is undefined or null, what should our expected output of this component be? do we want to render anything or just a message in this case, or default to course contentType?

src/components/catalogSearchResults/CatalogSearchResults.jsx
kiram15: I was planning on keeping the component preview because it gives the information that it is just the preview of courses/programs shown on the splitscreen page.

src/components/catalogs/CatalogSearch.jsx
kiram15: I feel like if contentType is undefined or null it should not render anything? Because that's a problem in our code, if we tried to render one and it wasn't correct it would error out in a way more gross way.

src/components/catalogSearchResults/CatalogSearchResults.jsx
binodpant: oh I did not realize it was an actual feature! if so yeah makes sense to keep it . I had mistaken it for a feature switch

src/components/catalogs/CatalogSearch.jsx
binodpant: yes I agree, if undefined, no point trying to render the rest, so maybe some default rendering is fine? What would we want to render then, just a message? or just null even? perhaps that is a product/UX question too, since that would be a case of bad data or something like that

src/components/catalogs/CatalogSearch.jsx
```diff
@@ -0,0 +1,89 @@
...
+import { useAlgoliaIndex } from './data/hooks';
+import PageWrapper from '../PageWrapper';
+
+import {
+  CONTENT_TYPE_COURSE, CONTENT_TYPE_PROGRAM, NUM_RESULTS_COURSE, NUM_RESULTS_PROGRAM, NUM_RESULTS_PER_PAGE,
+} from '../../constants';
+import CatalogSearchResults from '../catalogSearchResults/CatalogSearchResults';
+
+export default function CatalogSearch() {
+  const { refinements: { content_type: contentType, show_programs: programBool } } = useContext(SearchContext);
```
binodpant, line 16: how will we enable the show_programs feature when testing in staging or local dev, for example? will we need to feed it into the SearchData context?

```
<SearchData
        .... here?
      >
```

src/components/programCard/ProgramCard.jsx
```diff
@@ -16,33 +16,33 @@ import { injectIntl, intlShape } from '@edx/frontend-platform/i18n';
 import messages from './ProgramCard.messages';
 
 function makePlural(num, string) {
-  if (num > 1) { return (`${num} ${string}s`); }
+  if (num > 1 || num === 0) { return (`${num} ${string}s`); }
```
binodpant: great! 🎆

src/components/catalogSearchResults/CatalogSearchResults.jsx
```diff
@@ -246,6 +271,57 @@ export const BaseCatalogSearchResults = ({
...
+      Header: TABLE_HEADERS.numCourses,
+      accessor: 'course_keys',
+      Cell: ({ row }) => (row.values.course_keys.length > 0 ? `${row.values.course_keys.length}` : 'Available upon request'),
+    },
+    {
+      Header: TABLE_HEADERS.programType,
+      accessor: 'program_type',
+    },
+
+    // TODO: Badges commented out until Algolia bug is resolved (ENT-5338)
```
binodpant, line 299: cool thanks for the comments

binodpant (approved):
I am good to send this in once tests pass

PR #187, opened 2022-01-20
Feat: no results error and popular course display

alex-sheehan-edx:
Base no results expected behavior:

image

No results for courses but not programs:

image

 

# For all changes

### DIFF
--- a/src/components/catalogNoResultsDeck/CatalogNoResultsDeck.jsx
+++ b/src/components/catalogNoResultsDeck/CatalogNoResultsDeck.jsx
@@ -1,0 +1,114 @@
+import { Alert, CardView, DataTable } from '@edx/paragon';
+import React, { useEffect, useState } from 'react';
+import { injectIntl, intlShape } from '@edx/frontend-platform/i18n';
+import { logError } from '@edx/frontend-platform/logging';
+import PropTypes from 'prop-types';
+
+import {
+  CONTENT_TYPE_COURSE,
+  CONTENT_TYPE_PROGRAM,
+  NO_RESULTS_DECK_ITEM_COUNT,
+  NO_RESULTS_PAGE_SIZE,
+  NO_RESULTS_PAGE_ITEM_COUNT,
+} from '../../constants';
+import messages from './CatalogNoResultsDeck.messages';
+import EnterpriseCatalogApiService from '../../data/services/EnterpriseCatalogAPIService';
+import { getSelectedCatalogFromURL } from '../../utils';
+
+const BASE_APP_URL = process.env.BASE_URL;
+
+const CatalogNoResultsDeck = ({
+  intl,
+  setCardView,
+  columns,
+  renderCardComponent,
+  contentType,
+}) => {
+  const [defaultData, setDefaultData] = useState([]);
+  const [apiError, setApiError] = useState(false);
+
+  const selectedCatalog = getSelectedCatalogFromURL();
+  let redirect;
+  if (selectedCatalog) {
+    redirect = `${BASE_APP_URL}/?enterprise_catalog_query_titles=${encodeURIComponent(selectedCatalog)}`;
+  } else {
+    redirect = BASE_APP_URL;
+  }
+
+  useEffect(() => {
+    const defaultCoursesRefinements = { enterprise_catalog_query_titles: selectedCatalog, content_type: contentType };
+    EnterpriseCatalogApiService.fetchDefaultCoursesInCatalogWithFacets(defaultCoursesRefinements).then(response => {
+      setDefaultData(response.default_content);
+      setApiError(false);
+    }).catch(err => {
+      setApiError(true);
+      logError(err);
+      // TODO: what should the UX be for error here?
+    });
+  }, [selectedCatalog]);
+
+  let defaultDeckTitle;
+  let alertText;
+  if (contentType === CONTENT_TYPE_COURSE) {
+    alertText = intl.formatMessage(messages['catalogSearchResults.NoResultsCourseBannerText']);
+    defaultDeckTitle = intl.formatMessage(messages['catalogSearchResults.DefaultCourseDeckTitle']);
+  } else if (contentType === CONTENT_TYPE_PROGRAM) {
+    alertText = intl.formatMessage(messages['catalogSearchResults.NoResultsProgramBannerText']);
+    defaultDeckTitle = intl.formatMessage(messages['catalogSearchResults.DefaultProgramDeckTitle']);
+  }
+  return (
+    <>
+      <Alert className="mt-2" variant="info" data-testid="noResultsAlertTestId">
+        <Alert.Heading>{intl.formatMessage(messages['catalogSearchResults.NoResultsBannerTitle'])}</Alert.Heading>
+        {alertText}
+        <Alert.Link href={redirect}>
+          {intl.formatMessage(messages['catalogSearchResults.NoResultsBannerHyperlinkText'])}
+        </Alert.Link>
+      </Alert>
+      { (!apiError) && (
+        <h3 className="mt-4.5 mb-3.5" data-testid="noResultsDeckTitleTestId">{defaultDeckTitle}</h3>
+      )}
+      <DataTable
+        dataViewToggleOptions={{
+          isDataViewToggleEnabled: true,
+          onDataViewToggle: val => setCardView(val === 'card'),
+          togglePlacement: 'left',
+          defaultActiveStateValue: 'card',
+        }}
+        columns={columns}
+        data={defaultData}
+        itemCount={NO_RESULTS_DECK_ITEM_COUNT}
+        pageCount={NO_RESULTS_PAGE_ITEM_COUNT}
+        pageSize={NO_RESULTS_PAGE_SIZE}
+      >
+        <CardView
+          columnSizes={{
+            xs: 12,
+            sm: 6,
+            md: 4,
+            lg: 4,
+            xl: 3,
+          }}
+          CardComponent={(props) => renderCardComponent(props)}
+        />
+      </DataTable>
+    </>
+  );
+};
+
+CatalogNoResultsDeck.defaultProps = {
+  setCardView: () => {},
+  renderCardComponent: () => {},
+  columns: [],
+  contentType: '',
+};
+
+CatalogNoResultsDeck.propTypes = {
+  contentType: PropTypes.string,
+  intl: intlShape.isRequired,
+  setCardView: PropTypes.func,
+  renderCardComponent: PropTypes.func,
+  columns: PropTypes.arrayOf(PropTypes.shape({})),
+};
+
+export default injectIntl(CatalogNoResultsDeck);
--- a/src/components/catalogNoResultsDeck/CatalogNoResultsDeck.messages.js
+++ b/src/components/catalogNoResultsDeck/CatalogNoResultsDeck.messages.js
@@ -1,0 +1,36 @@
+import { defineMessages } from '@edx/frontend-platform/i18n';
+
+const messages = defineMessages({
+  'catalogSearchResults.DefaultCourseDeckTitle': {
+    id: 'catalogSearchResults.popularCourses',
+    defaultMessage: 'Popular Courses',
+    description: 'Popular Courses table header.',
+  },
+  'catalogSearchResults.DefaultProgramDeckTitle': {
+    id: 'catalogSearchResults.popularPrograms',
+    defaultMessage: 'Popular Programs',
+    description: 'Popular Programs table header.',
+  },
+  'catalogSearchResults.NoResultsBannerTitle': {
+    id: 'catalogSearchResults.NoResultsBannerTitle',
+    defaultMessage: 'No Results',
+    description: 'No results alert modal header.',
+  },
+  'catalogSearchResults.NoResultsCourseBannerText': {
+    id: 'catalogSearchResults.NoResultsCourseBannerText',
+    defaultMessage: 'No courses were found that match your search. Try ',
+    description: 'No results course alert modal text.',
+  },
+  'catalogSearchResults.NoResultsProgramBannerText': {
+    id: 'catalogSearchResults.NoResultsProgramBannerText',
+    defaultMessage: 'No programs were found that match your search. Try ',
+    description: 'No results programs alert modal text.',
+  },
+  'catalogSearchResults.NoResultsBannerHyperlinkText': {
+    id: 'catalogSearchResults.NoResultsBannerHyperlinkText',
+    defaultMessage: 'removing filters',
+    description: 'No results alert modal hyperlink text.',
+  },
+});
+
+export default messages;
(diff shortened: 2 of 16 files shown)


src/components/catalogSearchResults/CatalogSearchResults.test.jsx
```diff
@@ -152,18 +155,18 @@ describe('Main Catalogs view works as expected', () => {
...
-      <SearchDataWrapper>
-        <BaseCatalogSearchResults
-          {...defaultProps}
-          searchResults={emptySearchResults}
-        />
-      </SearchDataWrapper>,
-    );
-    expect(screen.queryByText(NO_DATA_MESSAGE)).toBeInTheDocument();
-  });
+  // test('no search results displays NO_DATA_MESSAGE', () => {
```
kiram15: I don't think we should comment this out, just remove it

src/data/services/EnterpriseCatalogAPIService.js
```diff
@@ -13,13 +13,26 @@ class EnterpriseCatalogApiService {
     return EnterpriseCatalogApiService.apiClient().get(enterpriseListUrl);
   }
 
+  static fetchDefaultCoursesInCatalog(options) {
```
kiram15: Should we add some tests for this file?

src/components/catalogNoResultsDeck/CatalogNoResultsDeck.test.jsx
```diff
@@ -0,0 +1,71 @@
...
+    formatTime: () => {},
+    formatRelative: () => {},
+    formatNumber: () => {},
+    formatPlural: () => {},
+    formatHTMLMessasge: () => {},
+    now: () => {},
+  },
+};
+
+describe('catalog no results deck works as expected', () => {
```
kiram15, line 70: Can we also add a test for removing filters?

kiram15 (changes requested):
Just some requests for more tests!

src/components/catalogNoResultsDeck/CatalogNoResultsDeck.jsx
```diff
@@ -0,0 +1,104 @@
...
+import React, { useState } from 'react';
+import { FormattedMessage, injectIntl, intlShape } from '@edx/frontend-platform/i18n';
+import { logError } from '@edx/frontend-platform/logging';
+import PropTypes from 'prop-types';
+
+import CourseCard from '../courseCard/CourseCard';
+import messages from './CatalogNoResultsDeck.messages';
+import EnterpriseCatalogApiService from '../../data/services/EnterpriseCatalogAPIService';
+
+export const NO_DATA_MESSAGE = 'No courses were found that match your search. Try ';
```
binodpant, line 11: Is this message complete?

src/components/catalogNoResultsDeck/CatalogNoResultsDeck.jsx
```diff
@@ -0,0 +1,104 @@
...
+  const [currentCatalog, setCurrentCatalog] = useState();
+
+  let redirect = process.env.BASE_URL;
+
+  const urlSearchParams = new URLSearchParams(window.location.search);
+  const params = Object.fromEntries(urlSearchParams.entries());
+  const selectedCatalog = params.enterprise_catalog_query_titles;
+
+  if (selectedCatalog) {
+    redirect += `/?enterprise_catalog_query_titles=${encodeURIComponent(selectedCatalog)}`;
```
binodpant: nit: maybe make the `redirect` a const outside the component like `BASE_APP_URL` and interpolate here instead like `${BASE_APP_URL}/?....` ?

src/components/catalogNoResultsDeck/CatalogNoResultsDeck.jsx
```diff
@@ -0,0 +1,104 @@
...
+  contentType,
+}) => {
+  const [defaultData, setDefaultData] = useState([]);
+  const [currentCatalog, setCurrentCatalog] = useState();
+
+  let redirect = process.env.BASE_URL;
+
+  const urlSearchParams = new URLSearchParams(window.location.search);
+  const params = Object.fromEntries(urlSearchParams.entries());
+  const selectedCatalog = params.enterprise_catalog_query_titles;
```
binodpant: This feels like it belongs in a util function somewhere.. even if it's only used here once.. chances are are already using this pattern elsewhere too, to read the query titles etc. It could even be a hook with useMemo if you like but at least a util function would be great I think, like getSelectedCatalogFromURL() or something like that

src/components/catalogNoResultsDeck/CatalogNoResultsDeck.jsx
```diff
@@ -0,0 +1,104 @@
...
+  }
+
+  // Request default courses only once unless the selected catalog has changed
+  if (!defaultData.length || currentCatalog !== selectedCatalog) {
+    if (currentCatalog !== selectedCatalog) {
+      setCurrentCatalog(selectedCatalog);
+    }
+
+    const defaultCoursesRefinements = { enterprise_catalog_query_titles: selectedCatalog, content_type: contentType };
+    EnterpriseCatalogApiService.fetchDefaultCoursesInCatalogWithFacets(defaultCoursesRefinements).then(response => {
```
binodpant, line 40: I think if you need to fetch data in a component code, you will want to use useEffect to ensure re-rendering does not re-invoke this, for example.

Ideally put this in a hook like `useNoResultsDefaultData()` or similar, use useEffect inside of that and call it here, would be my suggestion

as it stands, I believe it is going to re invoke the backend call on re-renders...

src/components/catalogNoResultsDeck/CatalogNoResultsDeck.jsx
```diff
@@ -0,0 +1,104 @@
...
+  contentType: '',
+};
+
+CatalogNoResultsDeck.propTypes = {
+  contentType: PropTypes.string,
+  intl: intlShape.isRequired,
+  cardClicked: PropTypes.func,
+  setCardView: PropTypes.func,
+  courseAssociatedCatalogs: PropTypes.arrayOf(PropTypes.string),
+  columns: PropTypes.arrayOf(PropTypes.object),
```
binodpant: if known/feasible, shall we try to further specify the PropTypes.shape({ }) for this columns prop?

src/components/catalogNoResultsDeck/CatalogNoResultsDeck.test.jsx
```diff
@@ -0,0 +1,71 @@
...
+      first_enrollable_paid_seat_price: 99,
+      original_image_url: '',
+      availability: ['Available Now'],
+    },
+  ],
+};
+
+// Enterprise catalog API mocks
+const mockGetCatalogApi = jest.spyOn(EnterpriseCatalogApiService, 'fetchDefaultCoursesInCatalogWithFacets');
+mockGetCatalogApi.mockResolvedValue(csvData);
```
binodpant: nit: perhaps better to do this in one statement?

```
const mockGetCatalogApi = jest.spyOn(
  EnterpriseCatalogApiService, 
  'fetchDefaultCoursesInCatalogWithFacets'
).mockResolvedValue(csvData);
```

src/components/catalogSearchResults/CatalogSearchResults.test.jsx
```diff
@@ -5,17 +5,20 @@ import '@testing-library/jest-dom/extend-expect';
...
 import messages from './CatalogSearchResults.messages';
 import { HIDE_PRICE_REFINEMENT } from '../../constants';
+import EnterpriseCatalogApiService from '../../data/services/EnterpriseCatalogAPIService';
 
 // Mocking this connected component so as not to have to mock the algolia Api
 const PAGINATE_ME = 'PAGINATE ME :)';
 const PaginationComponent = () => <div>{PAGINATE_ME}</div>;
 
+const csvData = [{ csv_data: 'foobar' }];
+const mockGetCatalogApi = jest.spyOn(EnterpriseCatalogApiService, 'fetchDefaultCoursesInCatalogWithFacets');
```
binodpant: ditto as this: https://github.com/openedx/frontend-app-enterprise-public-catalog/pull/187/files#r790989670

src/components/catalogSearchResults/CatalogSearchResults.test.jsx
alex-sheehan-edx: good catch, that was entirely cruft that I meant to remove hehe...

src/components/catalogNoResultsDeck/CatalogNoResultsDeck.jsx
alex-sheehan-edx: 👍 I'm fine with making this a util function

src/utils.js
```diff
@@ -0,0 +1,21 @@
+export function getSelectedCatalogFromURL() {
+  const urlSearchParams = new URLSearchParams(window.location.search);
+  const params = Object.fromEntries(urlSearchParams.entries());
+  return (params.enterprise_catalog_query_titles);
+}
+
+export function formatDate(courseRun) {
```
binodpant, line 7: love this utils file! I was going to do utils too so whoever merges later can refactor/merge

src/components/catalogSearchResults/CatalogSearchResults.jsx
```diff
@@ -144,21 +131,10 @@ export const BaseCatalogSearchResults = ({
...
-          description="Message is displayed when no data is returned (but no error occurs)."
-        />
-      </Alert>
-    );
-  }
-
   const { refinements, dispatch } = useContext(SearchContext);
   const nbHits = useNbHitsFromSearchResults(searchResults);
   const linkText = `Show (${nbHits}) >`;
+
```
binodpant, line 139: Just a thought, instead of conditionals later in the code, wouldn't it be simpler to return an altearnative markup/ cmponent here when SearchResults.nbHits === 0 ?

this way the rest of the code can prety much assume that nbHits > 0

Just wondering, to simplify potentially for less branching logic later in render code

alex-sheehan-edx: this was solved by the usage of `useEffect` 🥳

src/components/catalogNoResultsDeck/CatalogNoResultsDeck.messages.js
```diff
@@ -0,0 +1,31 @@
...
+    description: 'Popular Programs table header.',
+  },
+  'catalogSearchResults.NoResultsBannerTitle': {
+    id: 'catalogSearchResults.NoResultsBannerTitle',
+    defaultMessage: 'No Results',
+    description: 'No results alert modal header.',
+  },
+  'catalogSearchResults.NoResultsBannerText': {
+    id: 'catalogSearchResults.NoResultsBannerText',
+    defaultMessage: 'No courses were found that match your search. Try ',
```
kiram15, line 21: There should be two messages, one for courses and one for programs

alex-sheehan-edx: nice catch

src/components/catalogNoResultsDeck/CatalogNoResultsDeck.jsx
```diff
@@ -0,0 +1,111 @@
...
+      <DataTable
+        dataViewToggleOptions={{
+          isDataViewToggleEnabled: true,
+          onDataViewToggle: val => setCardView(val === 'card'),
+          togglePlacement: 'left',
+          defaultActiveStateValue: 'card',
+        }}
+        columns={columns}
+        data={defaultData}
+        itemCount={4}
```
kiram15: Should we put these in constants?

src/components/catalogSearchResults/CatalogSearchResults.jsx
```diff
@@ -382,9 +367,19 @@ export const BaseCatalogSearchResults = ({
...
+          <Button variant="link" onClick={() => refinementClick(contentType)}>{linkText}</Button>
+        )}
       </div>
       )}
+      { (searchResults?.nbHits === 0) && (
+        <CatalogNoResultsDeck
+          setCardView={setCardView}
+          columns={contentType === CONTENT_TYPE_COURSE ? courseColumns : programColumns}
+          renderCardComponent={renderCardComponent}
+          contentType={contentType === CONTENT_TYPE_COURSE ? 'course' : 'program'}
```
kiram15: NIT: I would use the constants here instead of string literals

src/utils.js
```diff
@@ -0,0 +1,21 @@
...
+    if (courseRun.start && courseRun.end) {
+      const startDate = (new Date(courseRun.start)).toLocaleDateString('default', { month: 'short', day: 'numeric', year: 'numeric' });
+      const endDate = (new Date(courseRun.end)).toLocaleDateString('default', { month: 'short', day: 'numeric', year: 'numeric' });
+      return `${startDate} - ${endDate}`;
+    }
+  }
+  return null;
+}
+
+export function makePlural(num, string) {
```
kiram15, line 18: This method could also be used here! https://github.com/openedx/frontend-app-enterprise-public-catalog/blob/16e7a72cad056183976f4b2a43a4954d29df89aa/src/components/programCard/ProgramCard.jsx#L18

src/components/catalogSearchResults/CatalogSearchResults.test.jsx
```diff
@@ -211,18 +224,6 @@ describe('Main Catalogs view works as expected', () => {
     );
     expect(screen.queryByText(PAGINATE_ME)).toBeInTheDocument();
   });
-  test('no search results displays NO_DATA_MESSAGE', () => {
```
kiram15, line 214: I think we need some tests for this: if search results = 0, we see the message on the screen, if there are courses but no programs, programs but no courses, etc

src/components/catalogSearchResults/CatalogSearchResults.test.jsx
```diff
@@ -150,6 +154,15 @@ const programProps = {
 };
 
 describe('Main Catalogs view works as expected', () => {
+  const OLD_ENV = process.env;
```
binodpant, line 157: this is good for now. I would really love to eliminate the use of process.env from our UI level code, and instead all of them just import from config. We can do a cleanup opportunistically instead of in a larger PR but this part looks like it's needed to handle the envs for now

alex-sheehan-edx: 👍  it was best practices and I noticed that this test didn't have the rest of old envs

src/components/catalogSearchResults/CatalogSearchResults.jsx
```diff
@@ -344,12 +331,23 @@ export const BaseCatalogSearchResults = ({
 
   function contentTitle() {
     let subTitle = (contentType === CONTENT_TYPE_COURSE) ? COURSE_TITLE : PROGRAM_TITLE;
-    if (refinements.q) {
+    if (refinements.q && refinements.q !== '') {
```
alex-sheehan-edx, line 334: I noticed a small bug where clicking the clear input field would result in a query param of an empty string. This would make this value display as `""` instead of nothing.

kiram15 (approved):
Looks good! 🚀 👍

PR #189, opened 2022-02-03
feat: Improve Formatting of CSV Download from Enterprise Catalog

johnnagro:
## Description

- use new workbook endpoint for download

## References

- [ENT-5381](https://openedx.atlassian.net/browse/ENT-5381)
- [ENT-5222](https://openedx.atlassian.net/browse/ENT-5222)


binodpant (approved):
lgtm! we should revisit testing but looks fine for this round of work

kiram15 (commented):
I know i've been a big stickler for tests recently, but we don't know when the next round of work will be. In the future, I would be wary of submitting things like this without testing even if we are under a deadline.

PR #190, opened 2022-02-03
fix: pulling content type value from array for comparison

alex-sheehan-edx:
# For all changes

### DIFF
--- a/src/components/catalogs/CatalogSearch.jsx
+++ b/src/components/catalogs/CatalogSearch.jsx
@@ -20,6 +20,14 @@ export default function CatalogSearch() {
   const showProgram = !!((programBool && programBool[0] === 'true'));
   const [noCourseResults, setNoCourseResults] = useState(false);
   const [noProgramResults, setNoProgramResults] = useState(false);
+
+  let specifiedContentType;
+  if (contentType) {
+    if (contentType.length === 1) {
+      [specifiedContentType] = contentType;
+    }
+  }
+
   return (
     <>
       <PageWrapper className="mt-3 mb-5 page-width">
@@ -86,7 +94,7 @@ export default function CatalogSearch() {
                   </Index>
                 </>
               )}
-              {(contentType === CONTENT_TYPE_PROGRAM) && (
+              {(specifiedContentType === CONTENT_TYPE_PROGRAM) && (
               <>
                 <Index indexName={algoliaIndexName} indexId="search-program">
                   <Configure
@@ -100,7 +108,7 @@ export default function CatalogSearch() {
               )}
             </>
             )}
-            {(contentType === CONTENT_TYPE_COURSE || !showProgram) && (
+            {(specifiedContentType === CONTENT_TYPE_COURSE || !showProgram) && (
             <>
               <Index indexName={algoliaIndexName} indexId="search-courses">
                 <Configure


src/components/catalogs/CatalogSearch.jsx
```diff
@@ -20,6 +20,14 @@ export default function CatalogSearch() {
   const showProgram = !!((programBool && programBool[0] === 'true'));
   const [noCourseResults, setNoCourseResults] = useState(false);
   const [noProgramResults, setNoProgramResults] = useState(false);
+
+  let specifiedContentType;
```
binodpant, line 24: is an array contentType supported in our app?

alex-sheehan-edx: what do you mean? Content type is an array because the user can select multiple content types to facet by

alex-sheehan-edx: image

binodpant: ouch okay so my PR also had optimizations to contentType which will need to adapt to this newfound knowledge then. Curious why we want to support multiple values for contenType, isn’t it going to be empty, course, or program, at any given time?

johnnagro: hmm yeah, as far as i understood - its "empty" on the main view, which means we see both. once you click "show" on a specific content type to see the course or program specific page, then you get a single content_type=<kind> param. i believe like binod, my download button makes those assumptions too

binodpant: aargh I am not thinking straight then. yes thank you for the correction. I was going to put content type handling in a util function so I can follow up once this fix is in, but yeah this looks good for this PR then! I think we want a quick hook like `useContentType` or `useAppParams` to fetch content type, that changes if the url param changes... or sth like it

src/components/catalogs/CatalogSearch.jsx
```diff
@@ -20,6 +20,14 @@ export default function CatalogSearch() {
...
+  if (contentType) {
+    if (contentType.length === 1) {
+      [specifiedContentType] = contentType;
+    }
```
binodpant, line 28: follow up: what happens in the case when content_type is size > 1?

src/components/catalogs/CatalogSearch.jsx
alex-sheehan-edx: I'm happy to pair on any refactors to the contentType handling!

src/components/catalogs/CatalogSearch.jsx
alex-sheehan-edx: as of rn, it's handled by this logic https://github.com/openedx/frontend-app-enterprise-public-catalog/blob/main/src/components/catalogs/CatalogSearch.jsx#L40-L88

alex-sheehan-edx: but I don't love that we have it hard coded to 2

src/components/catalogs/CatalogSearch.jsx
alex-sheehan-edx: However, since we demo'd this and it might get some hits soon, I think it's probably best if we merge this to get it working then can talk about improvements afterwards?

src/components/catalogs/CatalogSearch.jsx
binodpant: haha okay yeah thanks for clarifying. I now feel even stronger we need centralized handling of content type and logic around it. Since otherwise we will have to address similar stuff in multiple places. But I am good for this PR

binodpant (approved):
good for this fix, but yeah we should think of centralized handling for content type

src/components/catalogs/CatalogSearch.jsx
binodpant: agreed on quick fix.. with a later look at revisit of content type handling, should be a simple refactor PR we can do. Would prefer that to be post my work though :)

PR #192, opened 2022-02-11
feat: make image load whether it's a string or an object with a 'src' field

binodpant:
catalog server should have indexed image as string.. instead it currently does image: { src }

this change makes it handle both cases so we can render the program image correctly despite algolia state

we can undo this change once catalog server is updated to index image field as just a string instead

# For all changes


alex-sheehan-edx (approved):
lgtm

PR #194, opened 2022-02-14
fix: pass the correct facets in on the main view

johnnagro:
# Description

- when using the new programs-enabled-view, i noticed this error when trying to download from the main page

Screen Shot 2022-02-14 at 4 54 26 PM

- but it worked after clicking into a content type specific view via `show`
- I noticed a difference in the way `DownloadCsvButton` was passed `facets`

Screen Shot 2022-02-14 at 5 03 26 PM

- when making the outer one match the inner i got it working locally

## Testing

- tests pass
- local click-testing passed


kiram15 (approved):
LGTM, thank you for the quick turnaround

PR #195, opened 2022-02-15
fix: fixing list view modal bug

kiram15:
# For all changes

### DIFF
--- a/src/components/catalogModalBanner/CatalogCourseModalBanner.jsx
+++ b/src/components/catalogModalBanner/CatalogCourseModalBanner.jsx
@@ -39,9 +39,9 @@ const CatalogCourseModalBanner = ({
   endDate,
   upcomingRuns,
 }) => (
-  <div className="my-4.5 d-flex">
+  <div className="my-4.5 banner">
     <div className="banner-section mx-3">
-      <div className="d-flex h4 mb-0">
+      <div className="banner h4 mb-0">
         <Icon className="mr-1" src={MoneyOutline} />
         {coursePrice}
       </div>
@@ -53,7 +53,7 @@ const CatalogCourseModalBanner = ({
     <div className="banner-section slash">/</div>
     {checkSubscriptions(courseAssociatedCatalogs) && (
       <div className="banner-section mx-3">
-        <div className="d-flex h4 mb-0">
+        <div className="banner h4 mb-0">
           <Icon className="mr-1" src={BookOpen} />
           {intl.formatMessage(messages['CatalogCourseModalBanner.bannerCatalogText'])}
         </div>
@@ -64,7 +64,7 @@ const CatalogCourseModalBanner = ({
       <div className="banner-section slash">/</div>
     )}
     <div className="banner-section mx-3">
-      <div className="d-flex h4 mb-0">
+      <div className="banner h4 mb-0">
         <Icon className="mr-1" src={EventNote} />
         {checkAvailability(startDate, endDate)}
       </div>
--- a/src/components/catalogSearchResults/CatalogSearchResults.jsx
+++ b/src/components/catalogSearchResults/CatalogSearchResults.jsx
@@ -22,7 +22,7 @@ import {
   HIDE_PRICE_REFINEMENT,
   PROGRAM_TITLE,
 } from '../../constants';
-import { extractUuid, mapAlgoliaObjectToCourse, mapAlgoliaObjectToProgram } from '../../utils/algoliaUtils';
+import { mapAlgoliaObjectToCourse, mapAlgoliaObjectToProgram } from '../../utils/algoliaUtils';
 import CatalogInfoModal from '../catalogInfoModal/CatalogInfoModal';
 import { useSelectedCourse } from '../catalogs/data/hooks';
 import CourseCard from '../courseCard/CourseCard';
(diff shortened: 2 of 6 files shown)


src/components/catalogModalBanner/CatalogCourseModalBanner.jsx
```diff
@@ -64,7 +64,7 @@ const CatalogCourseModalBanner = ({
       <div className="banner-section slash">/</div>
     )}
     <div className="banner-section mx-3">
-      <div className="d-flex h4 mb-0">
+      <div className="banner h4 mb-0">
```
binodpant, line 67: guess I am not sure why this was not working, is d-flex not the same as doing display:flex?

src/components/catalogSearchResults/CatalogSearchResults.jsx
```diff
@@ -111,26 +111,12 @@ export const BaseCatalogSearchResults = ({
...
-      messages['catalogSearchResult.table.priceNotAvailable'],
-    );
     if (isProgramType) {
-      setSelectedCourse({
-        contentType: row.values.content_type,
-        programUuid: extractUuid(row.values.aggregation_key),
-        programTitle: row.values.title,
-        programProvider: row.values['partners[0].name'],
-      });
+      setSelectedCourse(
```
binodpant, line 115: hmm.. will this always work? I am not sure if I had missed this part but I thought the object passed to the `mapAlgoliaObjectToProgram` is expected to be an algolia style object, but some fields in the row.original may not fit that type/pattern

binodpant: e.g., row.values

src/components/catalogSearchResults/CatalogSearchResults.jsx
```diff
@@ -111,26 +111,12 @@ export const BaseCatalogSearchResults = ({
...
     } else {
-      setSelectedCourse({
-        ...row.original,
-        contentType: row.values.content_type,
-        price: priceText,
-        title: row.values.title,
-        skillNames: row.values.skill_names,
-        provider: row.values['partners[0].name'],
-      });
+      setSelectedCourse(mapAlgoliaObjectToCourse(row.original, intl, messages));
```
binodpant, line 119: yeah thanks for addressing this missing piece! but yes please check if the functions map*** will work.. you may have to write new functions since those may only work well with algolia input objects rn

kiram15: From my initial testing it seemed to work, is there a reason why it wouldn't? I'm using prod values to test

src/components/catalogModalBanner/CatalogCourseModalBanner.jsx
kiram15: If you look at enterprise_catalogs.scss, you can see that as the modal gets smaller, the display type has to change from flex to block in order to stack the components in a smaller window

src/components/catalogSearchResults/CatalogSearchResults.jsx
binodpant: hmm ok perhaps the fields line up just right then? I had thought that some values in row.value may or may not match what the function looks for

e.g., this is the expected input shape.. but yes as long as the row.values also meets this criterion we are fine, just wanted to make sure we double checked ... 

```
  const {
    title: courseTitle,
    partners,
    first_enrollable_paid_seat_price: coursePrice, // todo
    enterprise_catalog_query_titles: courseAssociatedCatalogs,
    full_description: courseDescription,
    original_image_url: bannerImageUrl,
    marketing_url: marketingUrl,
    advertised_course_run: courseRun,
    upcoming_course_runs: upcomingRuns,
    skill_names: skillNames,
  } = algoliaCourseObject;
```

binodpant: at my first attempt, I think I had found that not all the fields came from row.original but from row.values 


it used to look like this before my changes 


```
  const rowClicked = (row) => {
    const rowPrice = row.original.first_enrollable_paid_seat_price;
    const priceText = (rowPrice != null) ? `$${rowPrice.toString()}` : intl.formatMessage(
      messages['catalogSearchResult.table.priceNotAvailable'],
    );
    setPrice(priceText);
    setAssociatedCatalogs(row.original.enterprise_catalog_query_titles);
    setTitle(row.values.title);
    setProvider(row.values['partners[0].name']);
    setPartnerLogoImageUrl(row.original.partners[0].logo_image_url);
    setDescription(row.original.full_description);
    setBannerImageUrl(row.original.original_image_url);
    setMarketingUrl(row.original.marketing_url);
    setStartDate(row.original.advertised_course_run.start);
    setEndDate(row.original.advertised_course_run.end);
    setUpcomingRuns(row.original.upcoming_course_runs);
    setSkillNames(row.original.skill_names);
    open();
  };
```

binodpant:
also I don't think this is due to this PR but I noticed this error, perhaps while you are here we can fi xthis too?

```
index.js?1ef8:1 Warning: Each child in a list should have a unique "key" prop.

Check the render method of `ProgramModal`. See https://fb.me/react-warning-keys for more information.
    in CourseDisplayForProgram (created by ProgramModal)
```

src/components/catalogSearchResults/CatalogSearchResults.jsx
binodpant: yes awesome so this works for me too for both courses and programs once I switch to list view


so you can probably close this comment, just wanted to make sure we did due diligence

thanks

src/components/catalogModalBanner/CatalogCourseModalBanner.jsx
binodpant: ah ok 

```@media only screen and (max-width: 850px) {
  .slash {
    display: none;
  }
  .banner {
    display: block;
  }
```

hmm ok I see now, think this is fine for now. I dn't have a concrete thought but I remember seeing some trick somewhere which may have been briefer, but I can't remember rn.. thanks for explaining!

binodpant:
Did you also want to fix the `What you'll learn` piece that I had missed? (sorry again :) )

binodpant (approved):
lgtm!

PR #196, opened 2022-02-15
fix: remove significant trailing slash

johnnagro:
## Description

- changes in explore catalog routes require no trailing slash
- https://github.com/openedx/enterprise-catalog/pull/391

## Testing

- local click tests
- tested that prod works both ways


alex-sheehan-edx (approved):
lgtm

PR #199, opened 2022-02-28
fix: act warnings fix, and also fix proptypes error in CatalogNoResultsDeck

binodpant:
With node16 (coming up) many of these will become errors, due to this: https://developer.ibm.com/blogs/nodejs-15-release-blog

This PR unblocks the node16 PR: https://github.com/openedx/frontend-app-enterprise-public-catalog/pull/198

This PR fixes:

- [x] CatalogNoResultsDeck has a proptypes error since undefined data was being set. Instead I changed it to stay [] in this case. This fixes prop types warning (which is actually an error in node16)
- [x] fixes act warnings in the test overall by awaiting various elements in ui before returning. This also unblocks node16 (for this test at least)

# For all changes

### DIFF
--- a/src/components/catalogNoResultsDeck/CatalogNoResultsDeck.jsx
+++ b/src/components/catalogNoResultsDeck/CatalogNoResultsDeck.jsx
@@ -38,7 +38,7 @@ const CatalogNoResultsDeck = ({
   useEffect(() => {
     const defaultCoursesRefinements = { enterprise_catalog_query_titles: selectedCatalog, content_type: contentType };
     EnterpriseCatalogApiService.fetchDefaultCoursesInCatalogWithFacets(defaultCoursesRefinements).then(response => {
-      setDefaultData(response.default_content);
+      setDefaultData(response.default_content || []);
       setApiError(false);
     }).catch(err => {
       setApiError(true);
--- a/src/components/catalogSearchResults/CatalogSearchResults.test.jsx
+++ b/src/components/catalogSearchResults/CatalogSearchResults.test.jsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import { render, screen } from '@testing-library/react';
+import { act, render, screen } from '@testing-library/react';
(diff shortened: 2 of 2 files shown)


src/components/catalogNoResultsDeck/CatalogNoResultsDeck.jsx
```diff
@@ -38,7 +38,7 @@ const CatalogNoResultsDeck = ({
   useEffect(() => {
     const defaultCoursesRefinements = { enterprise_catalog_query_titles: selectedCatalog, content_type: contentType };
     EnterpriseCatalogApiService.fetchDefaultCoursesInCatalogWithFacets(defaultCoursesRefinements).then(response => {
-      setDefaultData(response.default_content);
+      setDefaultData(response.default_content || []);
```
binodpant, line 41: I figured keeping this as the default value of [] was fine, since in a test, we are somehow passing undefined to setDefaultData()

binodpant: this fixes the proptypes warning

binodpant: because DataTable has set `data` to be required

src/components/catalogSearchResults/CatalogSearchResults.test.jsx
```diff
@@ -188,6 +188,7 @@ describe('Main Catalogs view works as expected', () => {
     expect(screen.queryByText(TEST_PARTNER_2)).toBeInTheDocument();
 
     expect(screen.queryAllByText('A la carte').length === 2);
+    await act(() => screen.findByText('Business'));
```
binodpant, line 191: in general, the theme here is: wait for an async activity in UI to conclude before returning test, and wrap that wait in an act(). This simulates what a user experiences and also avoid failures in tests due to side effects that may run after tests return

alex-sheehan-edx (approved):
LGTM! ✅

PR #202, opened 2022-03-07
feat: adds default images for course/program cards

kiram15:
# For all changes

Screen Shot 2022-03-07 at 1 45 43 PM

Screen Shot 2022-03-07 at 1 48 45 PM


johnnagro (approved):

src/components/courseCard/CourseCard.jsx
```diff
@@ -31,6 +32,9 @@ const CourseCard = ({
         variant="top"
         src={card_image_url}
         alt={title}
+        onError={(e) => {
```
binodpant, line 35: lgtm! I saw some examples online using `event.onerror = null` after setting e.target.src but that is only needed if we feel like there can be an error after using e.target.src to load the image (not impossible but unlikely).

binodpant (approved):

PR #204, opened 2022-03-30
feat: implementing suggested search

alex-sheehan-edx:
# For all changes

### DIFF
--- a/src/components/catalogInfoModal/CatalogInfoModal.jsx
+++ b/src/components/catalogInfoModal/CatalogInfoModal.jsx
@@ -61,12 +61,16 @@ const CourseModal = ({
             </ModalDialog.Hero>
             <Image className="mr-2 partner-logo-thumbnail" src={partnerLogoImageUrl} rounded />
             <div className="padded-body">
-              <ModalDialog.Title className="h1 course-info-title">
-                {courseTitle}
-              </ModalDialog.Title>
-              <ModalDialog.Title className="h2 course-info-partner">
-                {courseProvider}
-              </ModalDialog.Title>
+              { courseTitle && (
+                <ModalDialog.Title className="h1 course-info-title">
+                  {courseTitle}
+                </ModalDialog.Title>
+              )}
+              { courseProvider && (
+                <ModalDialog.Title className="h2 course-info-partner">
+                  {courseProvider}
+                </ModalDialog.Title>
+              )}
               <CatalogCourseModalBanner
                 coursePrice={coursePrice}
                 courseAssociatedCatalogs={courseAssociatedCatalogs}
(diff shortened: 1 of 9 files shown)


src/components/catalogInfoModal/CatalogInfoModal.jsx
```diff
@@ -212,12 +216,16 @@ const ProgramModal = ({
             </ModalDialog.Hero>
             <Image className="mr-2 partner-logo-thumbnail" src={partnerLogoImageUrl} rounded />
             <div className="padded-body">
-              <ModalDialog.Title className="h1 course-info-title">
-                {programTitle}
-              </ModalDialog.Title>
-              <ModalDialog.Title className="h2 course-info-partner">
-                {programProvider}
-              </ModalDialog.Title>
+              { programTitle && (
```
alex-sheehan-edx, line 219: issue with course, program titles and providers being null on load and causing an error because modal dialog `title`s require children.

src/components/catalogSearchResults/CatalogSearchResults.jsx
```diff
@@ -342,7 +342,7 @@ export const BaseCatalogSearchResults = ({
           dataViewToggleOptions={toggleOptions}
           columns={isCourseType ? courseColumns : programColumns}
           data={tableData}
-          itemCount={searchResults?.nbHits}
+          itemCount={searchResults?.nbHits || 0}
```
alex-sheehan-edx, line 345: item count is required and from testing, occasionally missing, so I added a default

src/components/catalogPage/CatalogPage.test.jsx
```diff
@@ -43,4 +43,17 @@ describe('CatalogPage', () => {
     renderWithRouter(<CatalogPage />);
     expect(screen.getByText(selectionCardMessage['catalogSelectionDeck.edxForBusiness.label'].defaultMessage)).toBeInTheDocument();
   });
+  it('properly handles empty query params', async () => {
+    const location = {
```
binodpant, line 47: looks like async is not needed here?

src/components/catalogInfoModal/CatalogInfoModal.jsx
binodpant: alternative is to show skeleton or loading until all needed data is set, but this works too!

binodpant: I say it because if showing skeleton is acceptable ux wise it's a lot easier usually :)

src/components/catalogPage/CatalogPage.jsx
```diff
@@ -30,6 +30,13 @@ const CatalogPage = ({ intl }) => {
   const loadedSearchParams = new URLSearchParams(window.location.search);
   let reloadPage = false;
   let hideCards = false;
+
+  // Sometimes an empty query can be appended to the url, which can lead to confusing UX
```
binodpant, line 34: not sure I get why this is needed? will this make the page get reloaded?

alex-sheehan-edx: it will, I was noticing that occasionally the query params would end up `?q=` which was causing both visual and logic bugs so like how if other query params are missing/wrong we reload, I made it reload for this.

src/components/catalogs/CatalogSearch.test.jsx
```diff
@@ -37,4 +40,24 @@ describe('Catalog Search component', () => {
...
   });
+  it('properly renders component with program content type context', () => {
+    renderWithRouter(
+      <SearchDataWrapper
+        searchContextValue={COURSE_SEARCH_CONTEXT_VALUE}
+      >
+        <CatalogSearch />
+      </SearchDataWrapper>,
+    );
+    expect(screen.getByText('SEARCH')).toBeInTheDocument();
```
binodpant, line 51: this lgtm, just curious if there is any other ui piece that you can verify too, or is it basically the same ui for both cases?

src/components/catalogPage/CatalogPage.jsx
binodpant: ah ok sure, I just figured since we have an extra reload in the ui already not sure if this meant it would lead to an additional one. I don't know who/what appends the empty query to the url? do we know?

binodpant:
you will need to upgrade frontend-enterprise too in package.json

binodpant (approved):

PR #243, opened 2022-08-03
Update Explore Catalog dependencies

kiram15:
# For all changes


package.json
```diff
@@ -36,11 +35,13 @@
   "dependencies": {
     "@edx/brand": "npm:@edx/brand-openedx@1.1.0",
     "@edx/brand-openedx": "^1.1.0",
-    "@edx/frontend-component-footer": "10.2.4",
-    "@edx/frontend-component-header": "2.4.6",
-    "@edx/frontend-enterprise-catalog-search": "3.0.1",
-    "@edx/frontend-platform": "1.15.6",
-    "@edx/paragon": "19.18.0",
+    "@edx/frontend-build": "12.0.3",
```
adamstankiewicz: Let's keep `@edx/frontend-build` as a dev dependency, not a regular ol' dependency :)

package.json
```diff
@@ -61,18 +62,17 @@
     "react-helmet": "6.1.0",
     "react-instantsearch-dom": "6.23.3",
     "react-loading-skeleton": "2.2.0",
-    "react-redux": "5.1.2",
+    "react-redux": "7.2.6",
     "react-router": "5.3.1",
     "react-router-dom": "5.3.1",
-    "react-test-renderer": "17.0.2",
-    "react-use-hotjar": "1.2.1",
+    "react-test-renderer": "16.13.1",
```
adamstankiewicz, line 67: [curious] Is this downgrade of `react-test-renderer` necessary/intended?

src/components/catalogSearchResults/CatalogSearchResults.jsx
```diff
@@ -136,22 +115,57 @@ export const BaseCatalogSearchResults = ({
...
     if (isCourseType) { return <CourseCard {...props} onClick={cardClicked} />; }
     return <ProgramCard {...props} onClick={cardClicked} />;
-  }
+  };
+
+  const titleButton = useCallback((row) => (
+    <Button className="text-left" variant="link" onClick={() => rowClicked(row)}>
+      {row.values.title}
+    </Button>
+  ), [rowClicked]);
```
adamstankiewicz: nit: Rather than treating this as a function, it might be worth treating this as a React element/component instead. That is, define the component and its props outside of `CatalogSearchResults` itself and reference the element when passing it to `Cell`, e.g.:

```jsx
function TitleButton({ row, onClick }) {
  return (
    <Button className="text-left" variant="link" onClick={() => onClick(row)}>
      {row.values.title}
    </Button>
  );
}

function CatalogSearchResults() {
  return (
    <DataTable columns={[{ Cell: ({ row }) => <TitleButton {...row} onClick={rowClicked} /> }]},
  );
}
```

adamstankiewicz: The above suggestion would also apply to the other functions that are returning JSX, e.g. `catalogBadges`, `formatDate`, `titleButton`, etc. It may help to treat these are true functional React components than standard functions that return JSX.

package.json
kiram15: `1:26  error  '@edx/frontend-build' should be listed in the project's dependencies, not devDependencies  import/no-extraneous-dependencies`

Lint error that prompted the change, should I put in a comment to ignore it?

package.json
kiram15: Yes, so from what I understand, since we weren't as strict on these dependencies this wasn't caught earlier but this version of react-test-renderer needs at least react@"17.0.2", and because of a couple other dependencies that have restrictions on what version of react has to be used, this is what the error shows when building 

`npm ERR! code ERESOLVE
npm ERR! ERESOLVE unable to resolve dependency tree
npm ERR! 
npm ERR! While resolving: @edx/frontend-app-enterprise-public-catalog@1.0.0
npm ERR! Found: react@16.14.0
npm ERR! node_modules/react
npm ERR!   react@"^16.14.0" from the root project
npm ERR! 
npm ERR! Could not resolve dependency:
npm ERR! peer react@"17.0.2" from react-test-renderer@17.0.2
npm ERR! node_modules/react-test-renderer
npm ERR!   react-test-renderer@"^17.0.0" from the root project`

package.json
adamstankiewicz: Addressed in Slack :)

package.json
adamstankiewicz: Got it, makes sense. Thanks!

alex-sheehan-edx (approved):
I don't have all the context but this looks great from what I can tell 😄

PR #268, opened 2022-11-28
FEAT: refactoring catalog search to support exec ed content

alex-sheehan-edx:
enable changes by using `features=ENABLE_EXEC_ED_INCLUSION`

related PR: https://github.com/openedx/enterprise-catalog/pull/537

https://2u-internal.atlassian.net/browse/ENT-6383
https://2u-internal.atlassian.net/browse/ENT-6381
https://2u-internal.atlassian.net/browse/ENT-6577

inclusion of exec ed in catalog search + rework of structuring order of components rendered

Base landing page
image

Search results for only exec ed, not courses or programs
image

Search results for each learning type category
image

Search results for programs and courses but not exec ed
image

Exec ed specified learning type
image

# For all changes


src/components/catalogs/CatalogSearch.jsx
```diff
@@ -1,5 +1,5 @@
 /* eslint eqeqeq: "off" */
-import React, { useContext, useState, useMemo } from 'react';
+import React, { useContext, useState, useMemo, useEffect } from 'react';
```
kiram15: nit: alphabetize imports

src/components/catalogs/CatalogSearch.jsx
```diff
@@ -75,6 +99,51 @@ function CatalogSearch(intl) {
       setSelectedSuggestedCourseType('course');
     }
   };
+  
+  useEffect(() => {
+    // Grab content type(s) or default to course & program
+    let contentToDisplay = contentType ? contentType.slice() : [CONTENT_TYPE_COURSE, CONTENT_TYPE_PROGRAM];
+
+    // IFF the catalog query is a la carte, specified course type is exec ed (or course type isn't given)
```
kiram15: nit: two f's?

kiram15: Also this comment is confusing

src/constants.js
```diff
@@ -27,6 +27,7 @@ export const NUM_RESULTS_PROGRAM = 4;
 export const NUM_RESULTS_COURSE = 8;
 export const COURSE_TITLE = 'Courses';
 export const PROGRAM_TITLE = 'Programs';
+export const EXEC_ED_TITLE = 'Executive Education'
```
kiram15: nit: semi colon

src/components/catalogs/CatalogSearch.jsx
```diff
@@ -52,10 +74,12 @@ function CatalogSearch(intl) {
   }
 
   let suggestedSearchContentTypeFilter;
-  if (!contentType || contentType.length === 2) {
+  if (!contentType || contentType.length === 3) {
     suggestedSearchContentTypeFilter = combinedFilter;
   } else if (specifiedContentType === CONTENT_TYPE_PROGRAM) {
     suggestedSearchContentTypeFilter = programFilter;
+  } else if (specifiedContentType === CONTENT_TYPE_PROGRAM) {
```
kiram15: This is more just a point of clarification for me, aren't exec ed courses and not programs? also this if is identical to the one above it so you could just combine them if thats what you're trying to do. If not, this needs to be fixed

src/components/catalogs/CatalogSearch.jsx
```diff
@@ -75,6 +99,51 @@ function CatalogSearch(intl) {
       setSelectedSuggestedCourseType('course');
     }
   };
+  
+  useEffect(() => {
+    // Grab content type(s) or default to course & program
+    let contentToDisplay = contentType ? contentType.slice() : [CONTENT_TYPE_COURSE, CONTENT_TYPE_PROGRAM];
```
marlonkeating: nit: While slice() is a perfectly valid way to make a shallow copy, my general preference is to use an operator that does that specific operation, rather than a function built to do a different operation (but will do the operation you want if you omit certain arguments). 
```suggestion
    let contentToDisplay = contentType ? [...contentType] : [CONTENT_TYPE_COURSE, CONTENT_TYPE_PROGRAM];
```

src/components/catalogs/CatalogSearch.jsx
alex-sheehan-edx: IFF stands for "If and only if"

src/constants.js
alex-sheehan-edx: 🤦

src/components/catalogs/CatalogSearch.jsx
alex-sheehan-edx: love this!

src/components/catalogSearchResults/CatalogSearchResults.jsx
```diff
@@ -319,9 +388,13 @@ export function BaseCatalogSearchResults({
     );
   }
 
+  function CustomRowStatus() {
+    return null;
+  }
```
marlonkeating, line 408: clarification: What's the significance of 'null' for CustomRowStatus?  (As opposed to 'undefined' which is the other value used in line 458)

src/components/catalogSearchResults/CatalogSearchResults.jsx
```diff
@@ -260,7 +323,12 @@ export function BaseCatalogSearchResults({
     };
 
   function contentTitle() {
-    let subTitle = contentType === CONTENT_TYPE_COURSE ? COURSE_TITLE : PROGRAM_TITLE;
+    let subTitle;
```
alex-sheehan-edx, line 337: there's def a better way to do this

marlonkeating (commented):
I Don't see any large issues, but there's a few things I think we should consider as part of a future refactor (perhaps when we remove the flag gating).

To someone new to the code, it's quite hard to follow a lot of the logic between the particulars of the data table component, Algolia queries, and all the bits of filter logic (+ feature gating).  I think some additional compartmentalization and abstraction could greatly help readability and reusability.

src/components/catalogSearchResults/CatalogSearchResults.jsx
alex-sheehan-edx: oh uhhh, I figure undefined is preferable here

src/components/catalogSearchResults/CatalogSearchResults.jsx
```diff
@@ -208,6 +234,33 @@ export function BaseCatalogSearchResults({
...
+        Cell: TitleButtonComponent,
+      },
+      {
+        Header: TABLE_HEADERS.partner,
+        accessor: 'partners[0].name',
+      },
+      {
+        Header: TABLE_HEADERS.price,
+        accessor: 'entitlements',
+        Cell: ({ row }) => (row.values.entitlements[0].price
```
kiram15, line 251: The column version properly has price but the cards dont?

src/components/catalogSearchResults/CatalogSearchResults.jsx
kiram15: Honestly I would break it up into another if else, this way makes sense but its kind of confusing even if its less lines

kiram15 (changes requested):
Connected w/ Alex offine because there are some problems with some of the courses and programs not showing up as they should, which means changes need to be made before this can be merged in.

src/components/catalogSearchResults/CatalogSearchResults.jsx
alex-sheehan-edx: I just forgot to update the cards 🙈

src/components/catalogSearchResults/CatalogSearchResults.jsx
alex-sheehan-edx: actually, I'm taking that back, I believe this needs to be null.

kiram15 (approved):
LGTM!

PR #270, opened 2022-12-12
build: paragon version upgrade

Mashal-m:
`@edx/paragon` version was upgraded.


Mashal-m:
Closing because the paragon version is now already upgraded

PR #277, opened 2022-12-23
chore: upgrade @edx/paragon and use CardDeck.Deprecated

adamstankiewicz:
# For all changes

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,7 +40,7 @@
     "@edx/frontend-enterprise-catalog-search": "3.1.5",
     "@edx/frontend-enterprise-hotjar": "1.2.0",
     "@edx/frontend-platform": "2.5.1",
-    "@edx/paragon": "20.9.2",
+    "@edx/paragon": "20.25.0",
     "@fortawesome/fontawesome-svg-core": "1.2.35",
     "@fortawesome/free-brands-svg-icons": "5.15.3",
     "@fortawesome/free-regular-svg-icons": "5.15.3",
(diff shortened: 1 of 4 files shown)


src/components/catalogSelectionDeck/CatalogSelectionDeck.jsx
```diff
@@ -20,7 +20,9 @@ function CatalogSelectionDeck({ intl, title, hide }) {
     >
       <Container className="page-width">
         <h2>{title}</h2>
-        <CardDeck>
+        {/* TODO: The previously used `CardDeck` is since deprecated. However,
+        this use case is more relevant to `SelectableBox` than `CardDeck` */}
```
adamstankiewicz, line 24: https://paragon-openedx.netlify.app/components/selectablebox/

package.json
```diff
@@ -40,7 +40,7 @@
     "@edx/frontend-enterprise-catalog-search": "3.1.5",
     "@edx/frontend-enterprise-hotjar": "1.2.0",
     "@edx/frontend-platform": "2.5.1",
-    "@edx/paragon": "20.9.2",
+    "@edx/paragon": "20.25.0",
```
adamstankiewicz, line 43: Note: This upgrade repurposes the existing implementation of `CardDeck`, effectively as a breaking change. However, `CardDeck` was only used once throughout Open edX in this repo, so instead of releasing a breaking change just for a `CardDeck` refactor, we opted to release it as a new minor version and address the 1 usage of `CardDeck` through this PR.

That is, keep using the same exact one that is in use on production today, but renamed as `CardDeck.Deprecated`. I left a comment with a suggested component to use instead for the purpose of that UI use case.

src/components/catalogSelectionDeck/CatalogSelectionDeck.jsx
```diff
@@ -20,7 +20,9 @@ function CatalogSelectionDeck({ intl, title, hide }) {
...
+        <CardDeck.Deprecated>
```
adamstankiewicz, line 25: This is the the same as the previous `CardDeck` that was in use before, just renamed; no functional/style differences.

brobro10000 (approved):
LGTM

PR #281, opened 2023-01-18
fix: frontend fixes

kiram15:
https://2u-internal.atlassian.net/browse/ENT-6672
Catalog selection card fix (going to updated Paragon component SelectableBox)
Screen Shot 2023-01-18 at 5 28 48 PM
Download button no longer overlaps
Screen Shot 2023-01-18 at 5 29 03 PM
Page numbers center justified
Screen Shot 2023-01-18 at 5 29 42 PM
Text overlap in modal fix 
Screen Shot 2023-01-18 at 5 36 46 PM

# For all changes


src/components/catalogSelectionDeck/CatalogSelectionDeck.jsx
```diff
@@ -1,80 +1,83 @@
-import React from 'react';
+import React, { useContext, useEffect, useState } from 'react';
```
marlonkeating, line 1: nit: Import sorting

marlonkeating (approved):
Overall looks great!

src/components/catalogNoResultsDeck/CatalogNoResultsDeck.test.jsx
```diff
@@ -88,30 +84,34 @@ const execEdProps = {
...
+  test('it displays no results alert text', async () => {
     mockCatalogApiService.mockResolvedValue(csvData);
     render(
       <IntlProvider locale="en">
         <CatalogNoResultsDeck {...defaultProps} />
       </IntlProvider>,
     );
-    expect(screen.getByTestId('noResultsAlertTestId')).toBeInTheDocument();
-    expect(screen.getByText('No Results')).toBeInTheDocument();
+    await act(() => screen.findByTestId('noResultsAlertTestId'));
```
adamstankiewicz: Even though the tests warn you about using `act`, there are other approaches that are typically preferred over `act`. For example, you could try:

* `expect(await screen.findByTestId('noResultsAlertTestId'))`
* `await waitFor(() => { expect(screen.getByTestId('noResultsAlertTestId')); })`

See https://plainenglish.io/blog/you-probably-dont-need-act-in-your-react-tests-2a0bcd2ad65c for more details.

src/components/catalogNoResultsDeck/CatalogNoResultsDeck.test.jsx
```diff
@@ -16,10 +16,6 @@ const TEST_COURSE_NAME_2 = 'test course 2';
 const TEST_PARTNER_2 = 'edx 2';
 const TEST_CATALOGS_2 = ['baz', 'ayylmao'];
 
-// Note- console.error is causing tests to fail so we're simply getting rid of it :)
-delete global.console;
-global.console = { error: jest.fn() };
```
adamstankiewicz, line 21: 🙌

src/components/catalogSelectionDeck/CatalogSelectionDeck.jsx
```diff
@@ -1,80 +1,83 @@
...
-              messages['catalogSelectionDeck.edxForOnlineEdu.labelDetail'],
-            )}
-            cardBody={intl.formatMessage(
-              messages['catalogSelectionDeck.edxForOnlineEdu.body'],
-            )}
-          />
-        </CardDeck.Deprecated>
-      </Container>
-    </section>
+    <Container className="page-width" style={{ display: hide ? 'none' : 'block' }}>
```
adamstankiewicz: nit/curious: why is there a need for `.page-width` versus using the predefined `Container` `size` options: 

image

you could also opt to use `className={classNames({ 'd-none': hide })}` over inline style prop.

src/components/catalogSelectionDeck/CatalogSelectionDeck.jsx
```diff
@@ -1,80 +1,83 @@
...
+      <h2>{title}</h2>
+      <SelectableBox.Set
+        name="catalog-selection"
+        type="radio"
+        value={value}
+        onChange={handleChange}
+        columns={isExtraSmall ? 1 : 3}
+        className="py-4"
+      >
+        <SelectableBox value={config.EDX_ENTERPRISE_ALACARTE_TITLE} inputHidden={false} type="radio" aria-label="radio">
```
adamstankiewicz: nit: this may be a remnant of the Paragon docs site example, but each `SelectableBox` should have an `aria-label` that describes each individual `SelectableBox`. Currently, each `SelectableBox` has the same `aria-label="radio"` which to a screen reader isn't differentiating between them.

kiram15: I'll make this change! Also, would it be helpful if I also created a quick PR making the aria labels in Selectable Box unique for each instance in Paragon?

adamstankiewicz: @kiram15 If you have a few minutes to do so, that would be lovely :)

src/components/catalogSelectionDeck/CatalogSelectionDeck.jsx
kiram15: Honestly I think it was left in from when I wrote this and wasn't as familiar with Paragon, thanks for the catch!

PR #313, opened 2023-05-15
fix: wrapping catalog search results in paragon container to fit page formatting

alex-sheehan-edx:
With

image

Without

image

# For all changes

### DIFF
--- a/src/components/PageWrapper.jsx
+++ b/src/components/PageWrapper.jsx
@@ -7,8 +7,8 @@ import { PAGE_TITLE } from '../constants';
 export const DATA_TEST_ID = 'enterprise-catalogs-content';
 
 // eslint-disable-next-line react/prop-types
-const PageWrapper = ({ children, className }) => (
-  <Container className={className}>
+const PageWrapper = ({ children, className, size }) => (
+  <Container className={className} size={size}>
     <Helmet title={PAGE_TITLE} />
     <div data-testid={DATA_TEST_ID}>{children}</div>
   </Container>
@@ -16,6 +16,7 @@ const PageWrapper = ({ children, className }) => (
 
 PageWrapper.defaultProps = {
   className: '',
+  size: undefined,
 };
 
 export default PageWrapper;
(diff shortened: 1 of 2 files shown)


src/components/catalogPage/CatalogPage.jsx
```diff
@@ -139,7 +140,9 @@ const CatalogPage = ({ intl }) => {
             messages['catalogPage.catalogSelectionDeck.title'],
           )}
         />
-        <CatalogSearch />
+        <Container size="xl">
+          <CatalogSearch />
```
adamstankiewicz: While this works, you'll technically rendering a `Container` in a `Container` which in theory could lead to some unwanted behavior down the road.

Is there a way to update the `size` prop on the `Container` that already renders in `CatalogSearch` via its use of `PageWrapper`?

adamstankiewicz: https://github.com/openedx/frontend-app-enterprise-public-catalog/blob/de642b2bc604cd2ed656ab9c688f398cdab7e96a/src/components/PageWrapper.jsx#L11

alex-sheehan-edx: 👍

src/components/PageWrapper.jsx
```diff
@@ -7,15 +7,16 @@ import { PAGE_TITLE } from '../constants';
...
+const PageWrapper = ({ children, className, size }) => (
+  <Container className={className} size={size}>
     <Helmet title={PAGE_TITLE} />
     <div data-testid={DATA_TEST_ID}>{children}</div>
   </Container>
 );
 
 PageWrapper.defaultProps = {
   className: '',
+  size: '',
```
adamstankiewicz: I believe using empty string for prop to `Container` would result in a prop type warning:

`size` only accepts these types: https://github.com/openedx/paragon/blob/498357621207286f693bafedd6edf29762448dad/src/Container/index.jsx#L36

Use `undefined` here to match what `Container` expects.

adamstankiewicz (approved):

src/components/PageWrapper.jsx
alex-sheehan-edx: gooood catch

PR #319, opened 2023-06-08
fix: replace query-string with URLSearchParams

katrinan029:
## Description

**Problem:**
In `frontend-app-enterprise-public-catalog`, there are a few uses of `query-string`. This package has caused us lots of headache in the past due to us previously needing to stay pinned to v5 instead of latest, but also much of its functionality is natively supported by modern browsers via `URLSearchParams`.

The work was previously done to remove `query-string` from `frontend-app-learner-portal-enterprise`, `frontend-app-admin-portal`, and `frontend-enterprise`. But, we did not get to remove `query-string` uses from `frontend-app-enterprise-public-catalog`.

Related, `@edx/frontend-enterprise-utils` now exports a `hasFeatureFlagEnabled` utility function that can be re-used here in `frontend-app-enterprise-public-catalog` as well instead of re-defining a nearly identical function.

**Solution:** 
* Removes dependency on `query-string` in favor of `URLSearchParams`
* Re-adds `hasFeatureFlagEnabled` check to be able to temporarily turn on features on production for a given session.

Ticket: 
[Remove use of `query-string` in frontend-app-enterprise-public-catalog and use existing `hasFeatureFlagEnabled` utility function](https://2u-internal.atlassian.net/browse/ENT-5370)
# For all changes


src/config/index.js
```diff
@@ -1,24 +1,19 @@
...
 
-const hasFeatureFlagEnabled = (featureFlag) => {
-  const { features } = qs.parse(window.location.search);
-  return features && features.split(',').includes(featureFlag);
-};
-
 const features = {
   ENABLE_PROGRAMS:
-    process.env.FEATURE_ENABLE_PROGRAMS === 'true'
+    process.env.FEATURE_ENABLE_PROGRAMS
```
marlonkeating: If the environment variable ever equaled 'false', we'd still treat it as true here.  Is that what we want?

src/data/services/EnterpriseCatalogAPIService.js
```diff
@@ -9,16 +8,18 @@ class EnterpriseCatalogApiService {
...
+    }/catalog_workbook?${queryParams}${facetQuery}`;
     return enterpriseListUrl;
   }
 
   static fetchDefaultCoursesInCatalog(options) {
+    const queryParams = new URLSearchParams(options);
     const enterpriseListUrl = `${
       EnterpriseCatalogApiService.enterpriseCatalogServiceApiUrl
-    }/default_course_set?${qs.stringify(options)}`;
+    }/default_course_set?${queryParams.toString()}`;
```
marlonkeating, line 22: queryParams is already a string, per the return of createQueryParams

marlonkeating (changes requested):

src/config/index.js
katrinan029: Thanks for the catch! I overlooked that `process.env.FEATURE_ENABLE_PROGRAMS` was checking against a string rather than a boolean. I captured the changes in my latest commit.

src/data/services/EnterpriseCatalogAPIService.js
katrinan029: the `queryParams` here is not a result of calling the function `createQueryParams` since the options passed to `fetchDefaultCoursesInCatalog` do not contain an array values.

marlonkeating: Gotcha, missed that!  That said, template literal object expressions will have toString() called on them automatically, so technically there's no strict _need_ for toString() here.  That said, I kind of like seeing it explicitly laid out here.

marlonkeating (approved):

PR #378, opened 2024-03-12
[ENT-8322] feat: askXpert component to show user's query results and adding the ability to tweak them

mahamakifdar19:
**Jira Ticket**: https://2u-internal.atlassian.net/browse/ENT-8322
**Description**: Added a new react component to display askXpert query results to the user, allowing him to tweak the results by changing filter value from broad to focus.

**UI Screenshots**:
Screenshot 2024-03-13 at 3 12 11 PM
Screenshot 2024-03-13 at 3 12 23 PM

https://github.com/openedx/frontend-app-enterprise-public-catalog/assets/113524403/bf7771b0-9a95-4810-9cb8-0f711d969956

# For all changes


src/components/askXpertResultCard/AskXpertResultCard.jsx
```diff
@@ -0,0 +1,187 @@
...
+
+  const handleChange = (e) => {
+    const threshold = Number(e.target.value);
+    setThresholdValue(threshold);
+    fetchAICurationDataWithThreshold(threshold);
+  };
+
+  const selfPacedCoursesCount = items.filter(item => item.learning_type === 'course').length;
+  const selfPacedProgramsCount = items.filter(item => item.learning_type === 'program').length;
+  const execEdCoursesCount = items.filter(item => item.learning_type === 'Executive Education').length;
```
jajjibhai008: - We have already constants for these learning types. please use them
- Also can't we achieve this in a single filter iteration to make it efficient?

mahamakifdar19: @jajjibhai008 We need to maintain separate counts for different types, using multiple filters here is more appropriate IMO.

jajjibhai008: As we will get a large array of courses that's the reason I was saying to reduce it in a single iteration. 
btw you can consider this too if it looks good to you

```js
const counts = items.reduce((accumulator, item) => {
  switch(item.learning_type) {
    case 'course':
      accumulator.selfPacedCoursesCount++;
      break;
    case 'program':
      accumulator.selfPacedProgramsCount++;
      break;
    case 'Executive Education':
      accumulator.execEdCoursesCount++;
      break;
    default:
      break;
  }
  return accumulator;
}, { selfPacedCoursesCount: 0, selfPacedProgramsCount: 0, execEdCoursesCount: 0 });

const { selfPacedCoursesCount, selfPacedProgramsCount, execEdCoursesCount } = counts;
```

src/components/aiCuration/askXpertResultCard/AskXpertResultCard.jsx
```diff
@@ -0,0 +1,193 @@
...
+import { Close } from '@edx/paragon/icons';
+import askXpretImage from '../../../assets/edx-xpert-card-side-image.png';
+import EnterpriseCatalogAiCurationApiService from '../data/service';
+import {
+  CONTENT_TYPE_COURSE,
+  CONTENT_TYPE_PROGRAM,
+  EXEC_ED_TITLE,
+} from '../../../constants';
+
+const AskXpertResultCard = ({ taskId, queryTitle, data }) => {
```
saleem-latif: nit: I think `XpertResultCard` would be a better name for this.

src/components/aiCuration/askXpertResultCard/AskXpertResultCard.jsx
```diff
@@ -0,0 +1,193 @@
...
+  const [thresholdValue, setThresholdValue] = useState(0);
+  const [items, setItems] = useState(data);
+  const [loading, setLoading] = useState(false);
+
+  useEffect(() => {
+    setItems(data);
```
saleem-latif: Why do we need this?

src/components/aiCuration/askXpertResultCard/AskXpertResultCard.jsx
```diff
@@ -0,0 +1,193 @@
...
+  }, [data]);
+
+  const getXpertResultsWithThreshold = async (threshold) => {
```
saleem-latif: I think using a custom hook here would be better instead of JS function.

src/components/aiCuration/askXpertResultCard/AskXpertResultCard.jsx
```diff
@@ -0,0 +1,193 @@
...
+    try {
+      setLoading(true);
+      const results = await EnterpriseCatalogAiCurationApiService.getXpertResults(taskId, threshold);
+      const { status, data: responseData } = results;
+      if (status === 200 && responseData) {
```
saleem-latif: I think we should handle all HTTP status codes. 4XX and 5XX will be error and all other statuses can be considered success. 
More Info: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status

src/components/aiCuration/askXpertResultCard/AskXpertResultCard.jsx
```diff
@@ -0,0 +1,193 @@
...
+        setItems(responseData);
+      } else {
+        setItems([]);
```
saleem-latif: We should also check the error and developer message returned by the server here and handle accordingly.

src/components/aiCuration/askXpertResultCard/AskXpertResultCard.jsx
```diff
@@ -0,0 +1,193 @@
...
+                  </Form.Label>
+                  <datalist id="values" className="d-flex justify-content-between mb-1">
+                    <option value="0" label="Broad" className="p-0 font-weight-normal" />
+                    <option value="0.8" label="Focused" className="p-0 font-weight-normal" />
+                  </datalist>
+                  <Form.Control
+                    type="range"
+                    role="slider"
+                    value={thresholdValue}
+                    onChange={handleChange}
```
saleem-latif: We should use debouncing mechanism here. You can use https://lodash.com/docs#debounce this is one is already in the dependency list.

src/components/aiCuration/askXpertResultCard/AskXpertResultCard.jsx
```diff
@@ -0,0 +1,193 @@
...
+    }
+  };
+
+  const handleChange = (e) => {
+    const threshold = Number(e.target.value);
+    setThresholdValue(threshold);
+    getXpertResultsWithThreshold(threshold);
+  };
+
+  const selfPacedCoursesCount = items.filter(item => item.learning_type === CONTENT_TYPE_COURSE).length;
```
saleem-latif: I think it would be better to use `useMemo` here.

src/components/aiCuration/askXpertResultCard/AskXpertResultCard.jsx
```diff
@@ -0,0 +1,193 @@
...
+            <p className="text-white">Powered by OpenAI</p>
+          </Card.Section>
+        </Card>
+      </div>
+    </div>
+  );
+};
+
+AskXpertResultCard.defaultProps = {
+  taskId: null,
```
saleem-latif: I think `taskId` should not be optional, I do not think this component can function without `taskId`.

saleem-latif: I think `queryTitle` and `data` should also not be optional.

src/components/aiCuration/askXpertResultCard/AskXpertResultCard.jsx
```diff
@@ -0,0 +1,194 @@
...
+  queryTitle: null,
+  data: [],
+};
+
+AskXpertResultCard.propTypes = {
+  taskId: PropTypes.number,
+  queryTitle: PropTypes.string,
+  data: PropTypes.arrayOf(
```
saleem-latif: can we renamed `data` to something more readable?

src/components/aiCuration/askXpertResultCard/AskXpertResultCard.scss
```diff
@@ -0,0 +1,24 @@
+input[type='range'] {
+  accent-color: yellow;
+}
+
+input[type='range']::-webkit-slider-runnable-track {
+  width: 300px;
```
saleem-latif: I think we should avoid hardcoded values so that design is more responsive.

saleem-latif: Maybe you can use some classes from paragon? https://paragon-openedx.netlify.app/components/layout/

saleem-latif (changes requested):

src/components/aiCuration/askXpertResultCard/AskXpertResultCard.jsx
mahamakifdar19: @saleem-latif this is temporary and would be replaced by the attribute name returned by API response.

src/components/aiCuration/askXpertResultCard/AskXpertResultCard.scss
mahamakifdar19: @saleem-latif When styling the input[type='range'] element's slider track using Paragon classes, it's challenging to directly apply custom Paragon classes to override the default styling. This is because CSS rules targeting input[type='range']::-webkit-slider-runnable-track are specific to the browser's default styling.

saleem-latif (approved):

jajjibhai008 (approved):

PR #379, opened 2024-03-12
feat: create react component to all users to input querries and show …

jajjibhai008:
**Description**
Jira -> [ENT-8321](https://2u-internal.atlassian.net/browse/ENT-8321)

Initial state:
image

While Loading after posting Query:
image

Error State:
image

Whole Flow:

https://github.com/openedx/frontend-app-enterprise-public-catalog/assets/86868918/2f1934bc-be38-43c8-86a6-b63df741dd1b

# For all changes


src/components/aiCuration/AskXpert.jsx
```diff
@@ -0,0 +1,140 @@
...
+import LoadingBar from './ProgressBar';
+
+export const acceptableStatusCodes = [200, 201, 202, 204, 206, 207, 208, 226, 304];
+
+const AskXpert = ({ catalogName }) => {
+  const [isClose, setIsClose] = useState(false);
+  const [results, setResults] = useState([]);
+  const [isLoading, setIsLoading] = useState(false);
+  const [errorMessage, setErrorMessage] = useState('');
+  const onSubmit = async (query) => {
```
saleem-latif: nit: I think this would be a good candidate for a custom hook.

src/components/aiCuration/AskXpert.jsx
```diff
@@ -0,0 +1,140 @@
...
+    try {
+      setErrorMessage('');
+      setIsLoading(true);
+      const queryResponse = await EnterpriseCatalogAiCurationApiService.postXpertQuery(query, catalogName);
+      const { status: postRequestStatusCode, data: response } = queryResponse;
+      if (acceptableStatusCodes.includes(postRequestStatusCode)) {
```
saleem-latif: A better approach would be 
```
if (postRequestStatusCode >= 400) {
    // Error
} else {
    // Success
}
```

src/components/aiCuration/AskXpert.jsx
```diff
@@ -0,0 +1,140 @@
...
+        const resultsResponse = await EnterpriseCatalogAiCurationApiService.getXpertResults(response.task_id);
+        const { status: getRequestStatus, data: FinalResponse } = resultsResponse;
```
saleem-latif: nit: Use camelCase for constants. i.e.  use `finalResponse`.

src/components/aiCuration/AskXpert.jsx
saleem-latif: Actually 2 hooks would be better, one for the post request and another for polling.

src/components/aiCuration/AskXpert.jsx
```diff
@@ -0,0 +1,140 @@
...
+    } catch (error) {
+      setErrorMessage('An error occurred. Please try again.');
+    } finally {
+      setIsLoading(false);
+    }
+  };
+  // we will remove this line after passing the results to the results component
+  console.log('results', results);
+  const cancelSearch = () => {
+    setIsLoading(false);
```
saleem-latif, line 83: We should cancel the polling here as well.

src/components/aiCuration/AskXpert.jsx
```diff
@@ -0,0 +1,140 @@
...
+                  <Button variant="inverse-outline-primary" onClick={cancelSearch}>
+                    <FormattedMessage
+                      id="catalogPage.askXpert.cancel"
+                      defaultMessage="Cancel"
+                      description="Cancel button text for Xpert to cancel the search."
+                    />
+                  </Button>
+                </Col>
+                <Col xs={12} md={9}>
+                  <LoadingBar isLoading={isLoading} />
```
saleem-latif, line 139: We should be updating the loading bar based on the result of polling. i.e. increase bar as status changes from pending to in-progress to done/error.

saleem-latif: We will look into adding a better way to update the loading bar in the future.

src/components/aiCuration/AskXpert.jsx
```diff
@@ -0,0 +1,140 @@
...
+            </div>
+            )}
+          </Stack>
+        </Card.Section>
+        <Card.Section className="col-3 d-flex flex-column justify-content-between align-items-end pb-0">
+          <IconButton
+            invertColors
+            src={Close}
+            iconAs={Icon}
+            onClick={() => setIsClose(!isClose)}
```
saleem-latif, line 151: Can we use `useToggle` here? https://paragon-openedx.netlify.app/components/hooks/usetoggle/

src/components/aiCuration/AskXpertQueryField.jsx
```diff
@@ -0,0 +1,58 @@
...
+  IconButton,
+} from '@edx/paragon';
+import { Send } from '@edx/paragon/icons';
+import { useState } from 'react';
+
+import PropTypes from 'prop-types';
+
+const AskXpertQueryField = ({ onSubmit, isDisabled }) => {
+  const [textInputValue, setTextInputValue] = useState('');
+  const submitQuery = (value) => {
```
saleem-latif: I think we can skip the function definition and directly call `onSubmit` down there.

src/components/aiCuration/AskXpertQueryField.jsx
```diff
@@ -0,0 +1,58 @@
...
+  return (
+    <Form.Group>
+      <Form.Control
+        placeholder="Describe a skill, competency or job title you are trying to train"
+        onClick={(event) => { event.stopPropagation(); }}
+        onChange={(event) => { setTextInputValue(event.target.value); }}
+        disabled={isDisabled}
+        maxLength={300}
+        onKeyDown={(event) => {
+          if (event.key === 'Enter') {
```
saleem-latif: nit: Use `event. key === "Enter" || event. key === "NumpadEnter"`

src/components/aiCuration/data/service.js
```diff
@@ -0,0 +1,56 @@
...
+    }
+  }
+
+  static wait(ms) {
+    return new Promise(resolve => { setTimeout(resolve, ms); });
+  }
+
+  static async getXpertResults(taskId, retries = 0, threshold = 0) {
+    try {
+      const response = await axios.get(`${EnterpriseCatalogAiCurationApiService.enterpriseCatalogAiCurationServiceUrl}?task_id=${taskId}&&?threshold=${threshold}`);
```
saleem-latif, line 35: Here is an example to create a hook with polling https://www.dhiwise.com/post/a-guide-to-real-time-applications-with-react-polling 
We can enhance this example by clearing callback when status success/error is received indicating task completion.

saleem-latif (changes requested):

src/components/aiCuration/AskXpert.jsx
jajjibhai008: For now I can't implement this because we are not rendering this component in the parent. When we render this component in parent while integration then we can achieve this easily.

src/components/aiCuration/AskXpert.jsx
jajjibhai008: it some extra effort, and also since the blackened API is not final yet so it's a bit difficult to do this by just assuming things. As this is an extra step so we can achieve this after integration.

src/components/aiCuration/AskXpert.jsx
jajjibhai008: We can not create hooks there because we have to call those apis inside the onSumbit function and as per react rules we can not call a hook inside a function.

saleem-latif (approved):

mahamakifdar19 (approved):

PR #488, opened 2025-04-07
feat: upgrade react to v18

huniafatima-99:
# Description 
This PR upgrades react to v18 as part of [this epic](https://github.com/edx/edx-arch-experiments/issues/959). 

# For all changes

- [ ] Ensure adequate tests are in place (or reviewed existing tests cover changes)

PR #490, opened 2025-05-07
feat: add config for hooking localhost up to stage backend

marlonkeating:
[Jira Ticket](https://2u-internal.atlassian.net/browse/ENT-10369)

# Description

Persists the approach recommended by: https://2u-internal.atlassian.net/wiki/spaces/FEDX/pages/514162781/MFE+development+without+devstack

* Adds `.env.development-stage` file to persist the stage-specific configuration variables (e.g., pointing to stage APIs, stage JWT cookie names, etc.).
* Adds `webpack.dev-stage.config.js` file to persist the updated Webpack Dev Server configuration and load the environment variables from `.env.development-stage`.
* Adds NPM scripts `start:stage` to run the MFE locally against the `webpack.dev-stage.config.js` file, using environment variables defined in `.env.development-stage`. Allows more easily switching between local data vs. stage data.

## Usage

1. Define `ALGOLIA_APP_ID`, `ALGOLIA_SEARCH_API_KEY`, and `ALGOLIA_INDEX_NAME` for stage in the `.env.private` file (these should not be committed to the repository).
2. Execute the following command:
    * `npm run start:stage`
3. Load the app on https://localhost.stage.edx.org:8735/

# For all changes


package.json
```diff
@@ -16,6 +16,7 @@
     "lint:fix": "npm run lint -- --fix",
     "snapshot": "npm run test -- --updateSnapshot",
     "start": "fedx-scripts webpack-dev-server --progress",
+    "start:stage": "npm run start -- --config=webpack.dev-stage.config.js",
```
adamstankiewicz, line 19: [suggestion] I would also recommend adding `start:stage:with-theme` to be able to more easily run the app with the `@edx/brand-edx.org` theme (used on stage/prod) instead of the default Paragon theme.

```
"start:stage:with-theme": "paragon install-theme && npm run start:stage && npm install"
```

adamstankiewicz: [update] Technically, the `-edx` variants of the header/footer packages are installed, despite being in the `openedx` GitHub organization, so `npm start` and `npm run start:stage` does use `@edx/brand-edx.org` today; that said, these will be switch to the Open edX variants in the React 18 upgrade [PR](https://github.com/openedx/frontend-app-enterprise-public-catalog/pull/488#discussion_r2050731558) soon given its an `openedx` repo.

package.json
```diff
@@ -79,6 +81,8 @@
     "@testing-library/jest-dom": "5.17.0",
     "@testing-library/react": "16.3.0",
     "@testing-library/user-event": "14.6.1",
+    "dotenv": "8.6.0",
+    "dotenv-webpack": "3.0.0",
```
adamstankiewicz: [consideration] It might be worth just relying on these dependencies from `@openedx/frontend-build`, as transient dependencies to avoid having to worry about version mismatches between the MFE's `package.json` here and `@openedx/frontend-build`. For example, it looks like `@openedx/frontend-build` uses `dotenv-webpack@8` ([source](https://github.com/openedx/frontend-build/blob/e3856b0fab23dd271e6f018103f94bae36080421/package.json#L65)), where this is using a version 5 major releases behind.

I'm assuming these were added into `package.json` to appease the `no-extraneous-dependencies` ESLint error? I might recommend ignoring the ESLint rule vs. adding the dependencies. FWIW, the webpack config files in `frontend-app-learner-portal-enterprise` altogether ([source](https://github.com/openedx/frontend-app-learner-portal-enterprise/blob/6e05956f779dac803a7e02aa5938333333457da5/.eslintrc.js#L10)).

adamstankiewicz (approved):
LGTM! :shipit:

